PR #475, opened 2025-05-26
Allow device server to deal with lazy signals

### LINKED ISSUE #457: Devicemanager, handling ophyd devices with signals with `lazy=True`, i.e. `ADCpt`
## Bug report

The device server crashes and flushes the config for devices with `components` with `lazy=True` attributes if the signal can't connect. Explanation follows below. 

## Summary

The device manager creates device objects first, before it tries to connect to individual PVs. This logic was implemented to differentiate between issues to generate a class, and difficulties to connect, i.e. IOC host is not running. However, any signal that is `lazy=True` will try to connect if `hints` is called. This results in a crash within `get_device_info()` when the exception is raised and forwarded while the config is loaded (https://gitlab.psi.ch/bec/bec/-/blob/main/bec_server/bec_server/device_server/devices/devicemanager.py?ref_type=heads#L460).

The issue is within ophyd. Upon calling `hints` of a component with `lazy=True`, ophyd attempts to wait_for_signal to connect (https://github.com/bluesky/ophyd/blob/main/ophyd/device.py#L275). This is relevant for all `ADComponents` as they are lazy by default. 

For the moment, we can mitigate this behavior by setting the class attribute `lazy_wait_for_connection=False`. However, we would have to recursively set this on all sub-devices.

The following information may help to create a solution:

-> `get_device_info()` crashes here if the device has subsignal: https://gitlab.psi.ch/bec/bec/-/blob/main/bec_server/bec_server/device_server/devices/device_serializer.py?ref_type=heads#L164

## Expected Behavior vs Actual Behavior

The device_server should initial objects with signals `lazy=True`, without flushing the config in case the signals cannot be reached. The issue is linked to code in the upstream `ophyd` repository.

## Steps to Reproduce

The following test device crashes for `test_device.hints`, if the sub-device has a signal with `lazy=True`, but works for signals with `lazy=False` = default.

```python
from ophyd import Component as Cpt
from ophyd import Device, EpicsSignal, EpicsSignalRO, EpicsSignalWithRBV, Signal
from ophyd.areadetector import ADComponent as ADCpt

class TestDevice(Device):

    # lazy_wait_for_connection = False # Uncomment allows this code to work.

    sub_signal_ro = Cpt(EpicsSignalRO, "sub_signal_ro", kind="normal", lazy=True)


if __name__ == "__main__":
    # Example usage
    test_device = TestDevice(name="test_device")
    print(test_device)
    print(test_device.hints)
```
 Migrated from GitLab issue https://gitlab.psi.ch/bec/bec/-/issues/458

### LINKED ISSUE #476: Improve logging of Redis host during bec_ipython startup
If a beamline repository has a custom ServiceConfig with RedisServer HostPort, the BECIpythonClient would currently get stuck without any logging. A short log message should improve this

cappel89:
# Summary

This MR allows the device server to deal with lazy signals for device serialization, i.e. not try to connect to signals with `wait_for_connection()` if `.hints` is called. See issue #457 for more details. In addition, it improves logging of the Redis Host for bec_services to avoid a service being stuck due to not being able to connect to Redis because of a missconfiguration

# Related Issues

closes #457 
closes #476

### DIFF
--- a/bec_server/bec_server/device_server/devices/device_serializer.py
+++ b/bec_server/bec_server/device_server/devices/device_serializer.py
@@ -165,12 +165,9 @@ def get_device_info(
         # needed because ophyd signals have empty hints
         hints = {"fields": [obj.name]}
     else:
-        if not isinstance(obj, BECDeviceBase):
-            device_dict = get_devices_lazy_wait_for_connection(obj)
-            for _, info in device_dict.items():  # Set all to False
-                info[0].lazy_wait_for_connection = False
-        else:
-            device_dict = {}
+        device_dict = get_lazy_wait_for_connection(obj)
+        for _, info in device_dict.items():  # Set all to False
+            info[0].lazy_wait_for_connection = False
         hints = obj.hints
         for _, info in device_dict.items():  # Set back to initial value
             info[0].lazy_wait_for_connection = info[1]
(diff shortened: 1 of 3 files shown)


bec_server/bec_server/device_server/devices/device_serializer.py
```diff
@@ -161,7 +164,15 @@ def get_device_info(
         # needed because ophyd signals have empty hints
         hints = {"fields": [obj.name]}
     else:
+        if not isinstance(obj, BECDeviceBase):
```
wakonig, line 167: Probably better to move that logic into get_devices_lazy_wait_for_connection

bec_server/bec_server/device_server/devices/device_serializer.py
```diff
@@ -186,3 +197,32 @@ def get_device_info(
...
         },
     }
+
+
+def get_devices_lazy_wait_for_connection(
+    device: PositionerBase | ComputedSignal | Signal | Device | BECDeviceBase,
+    output: dict | None = None,
+) -> dict[str, tuple[PositionerBase | ComputedSignal | Signal | Device | BECDeviceBase, bool]]:
+    """
+    Get device and subdevices of a device.
```
wakonig, line 207: Please improve the doc string as it is not explaining correctly what the function is actually doing

wakonig: Maybe you mean 'Get lazy devices their subdevices'?

bec_server/bec_server/device_server/devices/device_serializer.py
```diff
@@ -186,3 +197,32 @@ def get_device_info(
...
+        dict[str, tuple[PositionerBase | ComputedSignal | Signal | Device | BECDeviceBase, bool]]: Dictionary of devices with their lazy_wait_for_connection status.
+    """
+    if output is None:
+        output = {}
+    if hasattr(device, "lazy_wait_for_connection"):
+        output[device.name] = (device, device.lazy_wait_for_connection)
+    else:
+        logger.warning(f"Device {device.name} does not have lazy_wait_for_connection attribute")
+
+    for attr, cpt in device._sig_attrs.items():  # pylint: disable=protected-access
```
wakonig, line 223: probably better to use the signal / component walk methods from ophyd rather than the internal variable

bec_server/bec_server/device_server/devices/device_serializer.py
cappel89: done

bec_server/bec_server/device_server/devices/device_serializer.py
cappel89: renamed!

PR #479, opened 2025-05-27
Add a CLI tool to generate widget plugins

d-perl:
Adds an application to manage creating plugin components in an installed plugin repository.

Imports the widget functionality from bec_widgets which can be merged separately in https://github.com/bec-project/bec_widgets/pull/718

Part of bec-project/bec_widgets#418


d-perl:
@wakonig thanks... I should have thought of that - they won't work until the bec_widgets part is merged since they are generated from the application

have moved them to a separate PR https://github.com/bec-project/bec/pull/532

wakonig (approved):

PR #484, opened 2025-05-27
fix(rpc): add obj to status base instances

wakonig:


bec_server/bec_server/device_server/device_server.py
```diff
@@ -187,7 +187,6 @@ def _update_instruction(self, instr_id: str) -> None:
         """
         request_info = self.get_request(instr_id)
         if request_info is None:
-            logger.warning(f"Received status object for unknown instruction {instr_id}.")
```
cappel89, line 190: Just wondering, why would we not need this logging?

cappel89 (approved):
lgtm

bec_server/bec_server/device_server/device_server.py
wakonig: because it is a valid case... if your status callback resolves before you insert it, it will trigger twice and completely misguide you in thinking that there is some leakage :D

PR #486, opened 2025-05-28
fix(test): dangling client thread

d-perl:
- makes sure the client is shutdown after procedure tests
- note that this should be handled by the procedure worker for any external process workers in the future, but not for the in-process worker which is tested here, since that would shut down the client for the process in the real BEC instance

### DIFF
--- a/bec_server/tests/tests_scan_server/test_procedures.py
+++ b/bec_server/tests/tests_scan_server/test_procedures.py
@@ -7,7 +7,7 @@ from unittest.mock import MagicMock, patch
 import fakeredis
 import pytest
 
-from bec_lib.client import RedisConnector
+from bec_lib.client import BECClient, RedisConnector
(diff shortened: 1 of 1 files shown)


bec_server/tests/tests_scan_server/test_procedures.py
```diff
@@ -32,6 +32,14 @@
 LOG_MSG_PROC_NAME = "log execution message args"
 
 
+@pytest.fixture(autouse=True)
```
cappel89, line 35: This is not leaking outside of this test module, correct? @wakonig @d-perl

cappel89 (commented):
lgtm, 
I would suggest to just have a quick check from @wakonig as I now that we've had issues with creating BECClient() instance within some scenarios as they are  singletons, and shutdown will permanently kill the logger for instance.

PR #492, opened 2025-06-02
fix(bec-yaml-loader): add support for relative filepaths; closes #483

### LINKED ISSUE #483: [BUG]: Yaml loader for device config should accept relative file paths

It should be possible to specify a relative file_path within the device_config via !include. This seems to crash currently:
https://gitlab.psi.ch/bec/debye_bec/-/jobs/390479

### Describe what you expected to happen and what actually happened.

It should accept relative file paths

### Outline the steps that lead to the bug's occurrence. Be specific and provide a clear sequence of actions.

_No response_

### bec_widgets version

_No response_

### bec core version

_No response_

### Any extra info / data? e.g. log output...

_No response_

### Related issues

_No response_

cappel89:
## Description

This PR fixes that !include statements in config files can include relative paths

## Related Issues

closes #483

## Definition of Done
- [x] Documentation is up-to-date.

### DIFF
--- a/bec_lib/bec_lib/bec_yaml_loader.py
+++ b/bec_lib/bec_lib/bec_yaml_loader.py
@@ -1,4 +1,5 @@
 import io
+from pathlib import Path
 
 import yaml
 
@@ -7,7 +8,10 @@ def include_constructor(loader, node):
     """
     Include another yaml file.
     """
-    filename = loader.construct_scalar(node)
+    filename = Path(loader.construct_scalar(node)).expanduser()
+    if not filename.is_absolute():
+        base_path = Path(loader.name).resolve().parent
+        filename = (base_path / filename).resolve(strict=False)
     with open(filename, "r", encoding="utf-8") as file_in:
         return {"__include__": {"data": yaml_load(file_in), "filename": filename}}
 
--- a/bec_lib/tests/test_yaml_loader.py
+++ b/bec_lib/tests/test_yaml_loader.py
@@ -84,9 +84,7 @@ def test_load_yaml_single_include_with_conflict(capfd, test_file1):
     assert len(out) == 1
     assert out["eiger"]["deviceClass"] == "ophyd_devices.SimCamera"
     out, _ = capfd.readouterr()
-    assert (
-        "Warning: Multiple definitions for key eiger. Using the one from ./test_file2.yaml." in out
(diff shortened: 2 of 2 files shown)


cappel89:
This bugfix should be merged quickly, blocks the merge of https://gitlab.psi.ch/bec/debye_bec/-/merge_requests/31

d-perl (approved):
lgtm!

bec_lib/bec_lib/bec_yaml_loader.py
```diff
@@ -8,6 +9,9 @@ def include_constructor(loader, node):
     Include another yaml file.
     """
     filename = loader.construct_scalar(node)
+    if not os.path.isabs(filename):
```
wakonig, line 12: "~/" would probably not work (albeit I haven't tested it)

bec_lib/bec_lib/bec_yaml_loader.py
```diff
@@ -7,7 +8,13 @@ def include_constructor(loader, node):
     """
     Include another yaml file.
     """
-    filename = loader.construct_scalar(node)
+    filename = Path(loader.construct_scalar(node)).expanduser()
+    if not filename.is_absolute():
+        base_path = Path(loader.name).resolve().parent
+        filename = (base_path / filename).resolve(strict=False)
+    if not filename.exists():
+        alternative_suffix = ".yml" if filename.suffix == ".yaml" else ".yaml"
```
wakonig, line 16: not the other way around? 
```
".yaml" if filename.suffix == ".yaml" else ".yml"
```

cappel89: No, so what I tried to accomplish is that you have a filename which is for example 'my_config.yml'

I then like to be able to write 'my_config.yaml', but it should still work. Probably there is a nicer way to accept both, yml/yaml as input for both extensions.

d-perl: I'm not sure, I think it would be better if it didn't look for files which were not specified, this makes me a bit nervous. At least I think it should log a warning.

cappel89: removed the 'smart' check as there could be 2 files.. Wrong filename --> raise

PR #506, opened 2025-06-05
refactor: enforce device config keys

d-perl:
- make `DeviceBase` run the config initialisation through a relaxed model to ensure all required keys are present
- update tests to match
- tidy tests a little


d-perl:
@wakonig the first commit is the actual changes - everything should pass there too, second commit is just me thinking the tests should be more parametrised and less copy paste

PR #512, opened 2025-06-05
refactor: expose more config handler function

d-perl:
- Split `send_config_request` into two parts to control waiting for the response, for use in BEC Widgets to deactivate the GUI while waiting
- Expose a list of non-updatable config keys

### DIFF
--- a/bec_lib/bec_lib/config_helper.py
+++ b/bec_lib/bec_lib/config_helper.py
@@ -10,6 +10,7 @@ import os
 import pathlib
 import time
 import uuid
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import yaml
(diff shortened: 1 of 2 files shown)


bec_lib/bec_lib/config_helper.py
```diff
@@ -33,6 +34,23 @@
 logger = bec_logger.logger
 
 
+@dataclass(frozen=True)
+class _ConfigConstants:
+    NON_UPDATABLE = ("name", "deviceClass", "deviceConfig", "description")
```
wakonig, line 39: * deviceConfig was updatable before (just handled before one would even get to this check)
* description should be updatable

wakonig (commented):
just some minor things... the description was indeed not updatable before. If I remember correctly, we added it after even having these checks. Or we simply forgot... either way, updating the description, especially from a GUI, could be quite interesting

bec_lib/bec_lib/config_helper.py
d-perl: fixed

wakonig (approved):

PR #516, opened 2025-06-08
feat(serializer): speed up serialization by avoiding ExtType; add new…

wakonig:
This PR fixes a problem in the serialization that caused the encoder to take roughly 2.5 x longer than normal. Moreover, I used the opportunity to refactor the serialization and make the conversion less random: The serialization now uses BECEncoder classes to register encoding and decoding methods. These encoders also provide the type information of objects that the encoder can translate. Any object of the specified type or subclasses will use the encoder. 

For now, the old serialization is still in the code base (serialization_registry_v1) and is used for backward compatibility. Once we have migrated BEC Widgets and redeployed BEC at the beamlines, we can remove the legacy section in serialization_registry and the entire serialization_registry_v1 module.

### DIFF
--- a/bec_lib/bec_lib/codecs.py
+++ b/bec_lib/bec_lib/codecs.py
@@ -1,0 +1,160 @@
+from __future__ import annotations
+
+import enum
+from abc import ABC, abstractmethod
+from typing import Any, Type
+
+import numpy as np
+from pydantic import BaseModel
+
+from bec_lib import messages as messages_module
+from bec_lib import numpy_encoder
+from bec_lib.device import DeviceBase
+from bec_lib.endpoints import EndpointInfo
+from bec_lib.messages import BECMessage, BECStatus
+
+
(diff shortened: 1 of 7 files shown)


d-perl:
Ah, I had some changes for the serialisation on my procedure worker branch: https://github.com/bec-project/bec/commit/ab50a2f4874043b62b539bf14b95af81c4b8f13c
If we are decoding to a pydantic model, it will accept any int for an int enum, so I thought it was better to encode them generically rather than special case every enum class, do you think you could make similar changes here?

wakonig:
Maybe I don't fully understand your comment, but https://github.com/bec-project/bec/pull/516/files#diff-24ab9916b35943e3725d668fafe04af1dcbb86fdf795ae2f91f8906dc7ed9fb9R78 should also decode to the enum with the new decoder

cappel89 (approved):
lgtm

PR #518, opened 2025-06-11
fix(logger): avoid re-initializing the connector when configuring the logger

wakonig:
## Description

[Provide a brief description of the changes introduced by this pull request.]

## Related Issues

[Cite any related issues or feature requests that are addressed or resolved by this pull request. Link the associated issue, for example, with `fixes #123` or `closes #123`.]

## Type of Change

- Change 1
- Change 2

## How to test

- Run unit tests
- Open [widget] in designer and play around with the properties

## Potential side effects

[Describe any potential side effects or risks of merging this PR.]

## Screenshots / GIFs (if applicable)

[Include any relevant screenshots or GIFs to showcase the changes made.]

## Additional Comments

[Add any additional comments or information that may be helpful for reviewers.]

## Definition of Done
- [ ] Documentation is up-to-date.

PR #520, opened 2025-06-12
fix(progress_signal): progress message must include scan id

wakonig:
This is just a quick fix to add the scan id to the progress message. Once we've migrated everything to use the bec signal, we can elevate the scan id to a proper pydantic field (cf #521). 

It also adds a fix for the initialization of the logger tests (introduced with #518).


cappel89 (approved):
lgtm, fyi. Do we have a scan that listens to a device progress at the moment?

PR #523, opened 2025-06-13
Fix/ensure unique device tags

d-perl:
changes `deviceTags` from a `list[str]` to a `set[str]`. `set_device_tags` still accepts any iterable so they can still be set with a list from the console.


d-perl:
@wakonig the pipelines are fine here, it just needs https://github.com/bec-project/bec_widgets/pull/702 as in the PR description

d-perl:
oh, haha, I didn't even notice the connector mock ones, because of the one where the set was encoded out of order

wakonig (approved):

wakonig (approved):

PR #530, opened 2025-06-18
Fix/signal update fix

### LINKED ISSUE #509: scan id access for async signal
Async signals rely on a scan id for their endpoint. However, at this point the scan id is fetched from the device metadata which could mean that data is pushed even after the scan was closed. Instead, these update methods should check if the scan is still running and only emit data if the scan has not been closed.

wakonig:
This PR should fix two related problems:
* Scan progress messages were set to 100 % whenever the scan finishes, even if it was aborted. In addition, the scan status was added to simplify the logic within UI components. 
* The message handler used to fetch metadata, such as the scan id directly from the device's metadata. This can lead to unexpected behavior and should rather be avoided. Instead, the data is now fetched directly from the DeviceManager's scan_info object. 

closes #509

### DIFF
--- a/bec_server/bec_server/device_server/bec_message_handler.py
+++ b/bec_server/bec_server/device_server/bec_message_handler.py
@@ -123,10 +123,22 @@ class BECMessageHandler:
         device_name = obj.root.name
         signal_name = obj.name
 
-        scan_id = self._get_current_scan_id()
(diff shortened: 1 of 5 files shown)


wyzula-jan (approved):
lgtm, works well with BEC Widgets requirements for progress bars.

PR #533, opened 2025-06-21
fix(serializer): reset codec cache when new codecs are registered

wakonig:


bec_lib/tests/test_serializer.py
```diff
@@ -177,3 +177,27 @@ def test_legacy_serialize(data):
...
+        def encode(obj):
+            return {"dummy": "data"}
+
+        @staticmethod
+        def decode(type_name: str, data: dict):
+            return DummyType()
+
+    assert msgpack.is_registered(DummyType) is False
+    msgpack.register_codec(DummyCodec)
+    assert msgpack.is_registered(DummyType) is True
```
d-perl, line 203: Do you prefer this syntax? I would have written:

```suggestion
    assert not msgpack.is_registered(DummyType)
    msgpack.register_codec(DummyCodec)
    assert msgpack.is_registered(DummyType)
```

because it is a bit superfluous to add another comparison when you already have a bool to assert, but it is a matter of taste what is more readable.

d-perl (approved):
lgtm!

d-perl (approved):

PR #537, opened 2025-06-24
fix(live_table): handle precision better

d-perl:
Tidy up the logic a little in printing the live table and handle bad inputs more gracefully


cappel89 (approved):
lgtm, briefly tested it too!

PR #540, opened 2025-06-26
feat(ipython): add support for negative precisions in epics

wakonig:
EPICS may report negative precisions to indicate a different formatting. This PR adds support to switch accordingly and also fixes the (valid) precision of 0.

### DIFF
--- a/bec_ipython_client/bec_ipython_client/callbacks/live_table.py
+++ b/bec_ipython_client/bec_ipython_client/callbacks/live_table.py
@@ -286,7 +286,7 @@ class LiveUpdatesTable(LiveUpdatesBase):
                         signals_precisions.append((None, None))
                     else:
                         prec = getattr(obj, "precision", 2)
-                        if not isinstance(prec, int) or prec < 1:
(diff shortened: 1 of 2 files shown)


bec_ipython_client/bec_ipython_client/callbacks/live_table.py
```diff
@@ -303,6 +303,10 @@ def _format_value(self, signal: dict | None, precision: int = 2):
             return "N/A"
         val = signal.get("value")
         if isinstance(val, SupportsFloat) and not isinstance(val, np.ndarray):
+            if precision < 0:
+                # This is to cover the special case when EPICS returns a negative precision.
+                # More info: https://epics.anl.gov/tech-talk/2004/msg00434.php
+                return f"{float(val):.{-precision}g}"
```
d-perl, line 309: Can this not just be collapsed to `{abs(precision)}` ?

wakonig: One uses "g", the other one "f" 😉

PR #543, opened 2025-06-30
Send scan_id and scan_number as alarms metadata

wakonig:
This PR adds the scan id and scan number to any alarm raised from the core services. This is only at best effort and there is no guarantee that the alarm will be sent with these metadata entries as we cannot know where in the code the error occurs. Yet, wrapping it in the core services simplifies the code in BEC Widgets. 

To this end, the scan base was slightly modified to accept the scan id per injection from the scan queue. Also the AsyncWriter received some minor changes.


wyzula-jan (approved):
Works well, thx

PR #549, opened 2025-07-10
fix(scan queue): remove stopped items from the queue

wakonig:
This PR ensures that stopped elements are removed from the queue. 

A bit of background: While in the past, only the first element could be stopped from the cli as only currently running scans could be halted, the GUI allows much more freedom, revealing a bug in the cleanup. If the queue item is set to stopped during the cleanup procedure, any subsequent insert of a new scan will block the entire server as it waits for the stopped element to be removed. 

Since we are not bombarding the backend with abort requests from a GUI, it is no surprise that we haven't observed the bug in our e2e tests.


cappel89 (approved):
Lgtm

PR #550, opened 2025-07-10
fix(scan history): filter scan history to readable files

### LINKED ISSUE #548: [BUG]: BEC History needs to check if file is accessible

The history does not check if a file is accessible for the account starting the session. This results in crashing clients that attempt to access files through bec.history, i.e. Waveform widget.

### Describe what you expected to happen and what actually happened.

_No response_

### Outline the steps that lead to the bug's occurrence. Be specific and provide a clear sequence of actions.

_No response_

### bec_widgets version

_No response_

### bec core version

_No response_

### Any extra info / data? e.g. log output...

_No response_

### Related issues

_No response_

wakonig:
The scan history needs to be filtered to readable items, otherwise the GUI accessing the scan history will raise immediately after changing accounts. 

closes #548

### DIFF
--- a/bec_lib/bec_lib/scan_history.py
+++ b/bec_lib/bec_lib/scan_history.py
@@ -4,6 +4,7 @@ This module contains the ScanHistory class, which is used to manage the scan his
 
 from __future__ import annotations
 
+import os
 import threading
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


wakonig:
It remains a mystery to me why os.access causes the pipeline to get stuck or massively slow down. I tried it with a custom basetemp path but that didn't help. Any input / ideas would be highly appreciated.

Locally it works just fine. I also tested that the same method on RHEL9 works. Maybe @d-perl can quickly test it on his machine. If it also works there, we can merge it as it is quite important to get this out.

wakonig:
Now, it works as expected. Thanks, @d-perl

PR #552, opened 2025-07-11
feat(ci): build and push requirements image

d-perl:
Add a step following the release to build an image containing a python env with the requirements for bec, and push it to ghcr

(I know there's a CI commit tag, but I would like there to be a release made with merging this)


d-perl:
ugh, I will close this and move the branch here for the tests

PR #557, opened 2025-07-14
fix(device config update): handle None

d-perl:
After https://github.com/bec-project/bec/pull/526 there were still some cases where updating a device with `{..., "deviceConfig": None}` would cause problems. Since we assume (despite the model) in many places that the actual `deviceConfig` can't be `None`, but we use `None` to mean "clear the config" in updates, as opposed to `{}` or omitting the item, which means "no update", I conservatively replace the existing config with an empty dict in this case. This corresponds to a user deleting the fields in the config editor and hitting submit.


d-perl:
closing this to handle this in the GUI for now, config logic needs bigger refactoring than we can do immediately

PR #565, opened 2025-07-22
fix: init devices with - separator

### LINKED ISSUE #564: Use `_child_name_separator = "-"` for our ophyd devices
To avoid ambiguity in readings (especially with nested devices with many dynamically created signals) we can take advantage of the changes introduced in https://github.com/bluesky/ophyd/pull/1222/

This is passed as an init argument to devices, but we would have to update anywhere we have hardcoded the underscore as part of keys in reading dictionaries

d-perl:
closes #564

### DIFF
--- a/bec_ipython_client/tests/end-2-end/test_scans_e2e.py
+++ b/bec_ipython_client/tests/end-2-end/test_scans_e2e.py
@@ -110,8 +110,8 @@ def test_mv_scan_nested_device(capsys, bec_ipython_client_fixture):
     if not bec.connector._messages_queue.empty():
         print("Waiting for messages to be processed")
         time.sleep(0.5)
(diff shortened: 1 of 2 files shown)


d-perl:
@wakonig what do you think of this? Doesn't look like it requires too many changes, but if we want to do it we should before anyone starts writing user scripts that access readings

d-perl:
Sounds good, I'll try to keep it up to date. @cappel89 @wyzula-jan try to write things that use the device `full_name` to access readings rather than the hardcoded underscore if you remember?

@wakonig should we start a v4 release branch soon then?

wakonig:
>@wakonig should we start a v4 release branch soon then?

Yes, if you want, you can already create pre_release_v4 branch and we merge it

PR #566, opened 2025-07-23
refactor: rename user scripts to user macros

wakonig:
With this PR, a few changes regarding the user scripts have been made:

* renaming to user_macros
* using composition instead of a mixin class
* add check to ensure that no code is execute upon loading the module 

We will have to update the beamline repos and inform the beamlines about the change.


d-perl (dismissed):
I think this is a good change, the wording makes more sense this way!

d-perl (approved):

PR #576, opened 2025-07-28
Improve DeviceServer device handling and logging

cappel89:
This PR improves 2 things:

(A)
It introduces a change for log message that will intercept any loggings from the 'ophyd' package and forwards them to the bec_logger. This will help identifying issues directly linked to ophyd where exceptions are not always properly attached to status objects, but only logged. One example for this is a failed move due to an alarm state (https://github.com/bluesky/ophyd/blob/main/ophyd/epics_motor.py#L304).

(B)
It fixes handling of the device serialization. Thereby, it checks that all signal_names are unique which is linked to issue #564. In addition, in case the device serialization fails for `connect=True` and `connect=False`, we now forward the proper error message

### DIFF
--- a/bec_server/bec_server/device_server/devices/device_serializer.py
+++ b/bec_server/bec_server/device_server/devices/device_serializer.py
@@ -130,16 +130,25 @@ def get_device_info(
     Returns:
         dict: updated device info
     """
-    protected_names = get_protected_class_methods()
+    # Check if the object namespace is valid
 
+    protected_names = get_protected_class_methods()
     user_access = get_custom_user_access_info(obj, {})
     if set(user_access.keys()) & set(protected_names):
         raise DeviceConfigError(
             f"User access method name {set(user_access.keys()) & set(protected_names)} is protected and cannot be used. Please rename the method."
         )
-
+    # Collect signals and their metadata
     signals = {}  # []
     if hasattr(obj, "component_names") and connect:
+        # Read attrs are only available if the device is connected, thus we can only deactivate the device
+        if hasattr(obj, "read_attrs"):
+            signal_names = [read_attr_name.replace(".", "_") for read_attr_name in obj.read_attrs]
+            unique_signal_names = set(signal_names)
+            if len(unique_signal_names) < len(signal_names):
+                raise DeviceConfigError(
+                    f"Signal names of {obj.name} must be unique, found duplicates. All signal names: {signal_names}. \n Unique signal names: {unique_signal_names}."
+                )
         walk = obj.walk_components()
         for _ancestor, component_name, comp in walk:
             if get_device_base_class(getattr(obj, component_name)) == "signal":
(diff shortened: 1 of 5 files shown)


bec_server/bec_server/device_server/devices/device_serializer.py
```diff
@@ -130,14 +130,23 @@ def get_device_info(
...
         raise DeviceConfigError(
             f"User access method name {set(user_access.keys()) & set(protected_names)} is protected and cannot be used. Please rename the method."
         )
 
+    if hasattr(obj, "read_attrs"):
+        signal_names = [read_attr_name.replace(".", "_") for read_attr_name in obj.read_attrs]
+        unique_signal_names = set(signal_names)
+        if len(unique_signal_names) < len(signal_names):
+            raise DeviceConfigError(
+                f"Signal names of {obj.name} must be unique, found duplicates. All signal names: {signal_names}. \n Unique signal names: {unique_signal_names}."
```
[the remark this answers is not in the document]
cappel89: disagree, we should use set for performance

d-perl (dismissed):
looks good!

bec_server/bec_server/device_server/devices/device_serializer.py
```diff
@@ -130,16 +130,25 @@ def get_device_info(
...
         raise DeviceConfigError(
             f"User access method name {set(user_access.keys()) & set(protected_names)} is protected and cannot be used. Please rename the method."
         )
-
+    # Collect signals and their metadata
     signals = {}  # []
     if hasattr(obj, "component_names") and connect:
+        # Read attrs are only available if the device is connected, thus we can only deactivate the device
+        if hasattr(obj, "read_attrs"):
+            signal_names = [read_attr_name.replace(".", "_") for read_attr_name in obj.read_attrs]
```
wakonig, line 146: shouldn't `describe` give you the same?

cappel89: Describe is not sufficient, but neither did the above properly work. Sorry, it got lost when I moved things around.  I pushed a fix that captures it now! Please check again @wakonig

wakonig: why would describe not be sufficient? it has all read signals, no?

cappel89: That is exactly the point, it overwrites whatever is conflicting with the name.

```
class SynDeviceOPAAS(Device):
    x = Cpt(SimPositioner, name="x")
    y = Cpt(SimPositioner, name="y")
    z = Cpt(SynDeviceSubOPAAS, name="z")
    x_setpoint = Cpt(SimPositioner, name="x_setpoint")

hexapod = SynDeviceOPAAS(name="hexapod')
```
Here the hexapod.describe() only has a single entry for `hexapod_x_setpoint`

wakonig (approved):

PR #593, opened 2025-09-01
feat(script executor): improve message interface for aborted scripts

wakonig:
This PR includes minor improvements to the script executor:

* Add default ttl for scripts. Previously, scripts were automatically cleaned up after execution but for development it is better to clean up after some time. 
* Add traceback info to `ScriptExecutionInfoMessage`. This will allows us to send error messages to the IDE. The traceback frames are trimmed to the actual execution. 
* Add "aborted" exit status and fix KeyboardInterrupt handling


d-perl (approved):
lgtm, I like the nice cleaned up traceback!

PR #605, opened 2025-09-12
fix: rework signature serializer

d-perl:
signature serializer is quite a lot more robust; will just drop any types it doesn't know about

currently requires python 3.11+

### DIFF
--- a/bec_server/bec_server/data_processing/lmfit1d_service.py
+++ b/bec_server/bec_server/data_processing/lmfit1d_service.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import inspect
 import threading
 import time
-from typing import TYPE_CHECKING
 
 import lmfit
 import numpy as np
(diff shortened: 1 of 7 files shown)


bec_server/bec_server/scan_server/scans.py
```diff
@@ -15,14 +15,12 @@
 from bec_lib.devicemanager import DeviceManagerBase
 from bec_lib.endpoints import MessageEndpoints
 from bec_lib.logger import bec_logger
+from bec_server.scan_server.instruction_handler import InstructionHandler
```
wakonig, line 18: the import is only needed for type hints, no?

bec_server/bec_server/data_processing/lmfit1d_service.py
```diff
@@ -13,11 +12,9 @@
 from bec_lib.endpoints import MessageEndpoints
 from bec_lib.lmfit_serializer import deserialize_param_object, serialize_lmfit_params
 from bec_lib.logger import bec_logger
+from bec_lib.scan_items import ScanItem
```
wakonig, line 15: the import is only needed for type hints

PR #620, opened 2025-10-06
fix(device): update default cached behavior for read methods to False

wakonig:
Now that read operations can bypass the queue, we can set the default to cached=False. 

https://github.com/bec-project/bec_widgets/pull/869 needs to be merged first


wakonig:
> https://github.com/bec-project/bec_widgets/pull/869 needs to be merged first

;)

PR #622, opened 2025-10-08
fix(history): fix cache for h5 structure access

wakonig:
The structure of h5 files was only stored inside the scan data container object, thus not available to new objects pointing to the same file. 

This PR introduces an LRU cache for the file structure and caches the last 100 file structures. 

It also removes pydantic 2.12.0 as compatible version. Let's see if they can fix it with a new patch release. Looks like we are not the only ones having problems.

### DIFF
--- a/bec_lib/pyproject.toml
+++ b/bec_lib/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "msgpack~=1.0, >1.0.4",
     "numpy>=1.24, <3.0",
     "psutil~=5.9",
-    "pydantic~=2.8",
+    "pydantic~=2.8, !=2.12.0",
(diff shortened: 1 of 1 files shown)


bec_lib/bec_lib/scan_data_container.py
```diff
@@ -147,6 +188,7 @@ def read(self, entry_path: str, cached=True, entry_filter: list[str] | None = No
         out = {}
         with h5py.File(self.file_path, "r") as f:
             entry = f[entry_path]
+            logger.info(f"Reading entry {entry_path} from file {self.file_path}")
```
wyzula-jan, line 191: do we want to keep this?

wyzula-jan (approved):

PR #641, opened 2025-10-30
fix(macro_update_handler): avoid shadowing the imported module

wakonig:
Fix to support reloading single macros from a module without shadowing already imported modules. I'm not a big fan of the solution, so if anyone has a better idea, I'd be open for it.


cappel89 (approved):
lg

PR #644, opened 2025-11-10
fix(rpc): update config cache on set commands

wakonig:
This PR fixes a bug in the rpc handling on the device server where .set commands would not lead to an update of the config cache in redis. 

Moreover, the RPCMixin class has been refactored to a standalone component: `RPCHandler`. The object retrieval has been unified. Private methods of the DeviceServer class needed by the RPCHandler are now public.


bec_server/bec_server/device_server/rpc_mixin.py
```diff
@@ -156,6 +156,25 @@ def _handle_misc_rpc(self, instr: messages.DeviceInstructionMessage) -> Any:
                 instr_params.get("func").split(".put")[0],
             )
             self._update_cache(obj, instr)
+        elif instr_params.get("func") == "set" or instr_params.get("func").endswith(".set"):
+            if instr_params.get("func") == "set":
+                obj = self.device_manager.devices[device_root].obj
+            else:
```
d-perl: This is a little hard to read, and it might be nicer in its own function rather than "misc" RPC? this whole thing could use a little refactoring with a helper func to get the correct device obj since that is repeated quite a few times here for the different operations

d-perl (approved):
This looks like it works so I think it's okay, but I would probably prefer if it was tidied up a little

PR #645, opened 2025-11-11
Refactor/config handler

wakonig:
This PR fixes an issue where a config update instruction would crash when the device does not exist. As a quick fix, the device is simply skipped. I will introduce an improved update routine in a follow-up PR. 

Moreover, I re-ordered the methods and added doc strings.

### DIFF
--- a/bec_server/bec_server/scihub/atlas/config_handler.py
+++ b/bec_server/bec_server/scihub/atlas/config_handler.py
@@ -106,10 +106,20 @@ class ConfigHandler:
             self.send_config_request_reply(accepted=True, error_msg=None, metadata=msg.metadata)
 
     def _reload_config(self, msg: messages.DeviceConfigMessage):
+        """
+        Reload the config in all services.
+        Args:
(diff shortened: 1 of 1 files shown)


bec_server/bec_server/scihub/atlas/config_handler.py
```diff
@@ -73,7 +75,51 @@ def send_config_request_reply(self, accepted, error_msg, metadata):
...
+        Args:
+            msg (messages.DeviceConfigMessage): Config update message
+
+        """
+        updated = False
+        dev_configs = msg.content["config"]
+
+        for dev, config in dev_configs.items():
+            if dev not in self.device_manager.devices:
+                continue
```
cappel89: maybe a brief comment, pointing people to https://github.com/bec-project/bec/blob/main/bec_lib/bec_lib/devicemanager.py#L691

wakonig: I will anyway refactor the module soon... the PR was simply to track the changes more easily

PR #647, opened 2025-11-12
Feat/add procedure api

d-perl:
Adds functionality to the procedure manager and associated tooling:
- a helper class for Redis interactions, separated for frontend and backend use
- get statuses of running procedures
- abort procedures, reschedule aborted procedures, delete aborted procedures
- new messages for the above
- a builtin procedure for running user scripts
- tests

Helpful for testing:
- run the `_dev_runner.py`
- Make sure you have a pod created with `podman pod create --add-host redis:host-gateway local_bec`

- In a bec console:
```python
from bec_server.scan_server.procedures.helper import BackendProcedureHelper
from bec_lib import messages
helper = BackendProcedureHelper(bec.connector)
msg = messages.ProcedureRequestMessage( identifier="sleep", queue="queue1", args_kwargs=((), {"time_s":5}) )
helper.request_procedure(msg)
```

https://github.com/bec-project/bec_widgets/pull/902 can be used to monitor queues, abort and reschedule procedures, as well as run scripts as procedures

### DIFF
--- a/bec_server/tests/tests_scan_server/test_procedures.py
+++ b/bec_server/tests/tests_scan_server/test_procedures.py
@@ -40,6 +40,7 @@ FAKEREDIS_PORT = 6380
 
 
 @pytest.fixture(autouse=True)
+@patch("bec_lib.bec_service.BECAccess", MagicMock)
 def shutdown_client():
     bec_client = BECClient(
         config=ServiceConfig(config={"redis": {"host": FAKEREDIS_HOST, "port": FAKEREDIS_PORT}}),


bec_lib/bec_lib/devicemanager.py
```diff
@@ -650,16 +650,16 @@ def _add_device(self, dev: dict, msg: DeviceInfoMessage):
         class_name = info["device_info"]["device_class"]
 
         if base_class == "device":
-            logger.info(f"Adding new device {name}")
+            logger.debug(f"Adding new device {name}")
```
wakonig: I'm not sure about the change of log level. On one hand, it greatly reduces the spamming. On the other hand, it can be very useful whilst trying to see some progress when devices are slow to connect. But maybe we remove it for now and if we really can't live without it, we add it again? Or we replace the direct logs with just a summary of devices that have been added?

d-perl:
@wakonig I just undid the logging commit, we should maybe discuss it in more detail sometime outside of this PR

wakonig:
just out of curiosity... could you rebase your branch on main and see if the coverage report is fixed? Also fine if you don't want to do this on this branch ;)

PR #648, opened 2025-11-12
device server shutdown

wakonig:
This PR modifies the device server to disconnect from all devices on shutdown. Moreover, the tmux wrapper has been modified to send a SIGINT signal to gracefully shut down the services. If the SIGINT does not work, the session is killed after a some time (default: 5 s). 

It also removes the class variables from the device manager base. This makes the tests much more predictable.


wakonig:
> looks good, just fix the commit title?

Thank you! - I completely forgot to rebase

d-perl (approved):

PR #657, opened 2025-11-22
Fix/file writer

wakonig:
This PR modifies the file writer to write to .tmp files first before renaming them to .h5. While it doesn't fully prevent access to files while they are written, it at least makes it less likely.


cappel89 (approved):
looks good.

PR #669, opened 2025-11-29
test: disabling script executor

wakonig:
There are issues with coverage reporting when using the script executor in CI. Until these are resolved, we are disabling it to ensure accurate coverage metrics.

Related issue: #670


d-perl (dismissed):
This is different from the other coverage problems we've noticed, right? 

Anyway, I'd prefer to use `pytest.mark.skip` over commenting them out, but either way is fine

PR #676, opened 2025-12-03
Feature/beamline states

### LINKED ISSUE #406: Move `beamline checks` from the client to the server
`bl_checks` and `bl_conditions` provide for constantly-monitored checks on beamline state to interrupt, invalidate, or repeat measurements if values go out of bounds. Currently these are implemented in the iPython client. Instead, they should be handled in the scan server.

Each check, as registered with the server, should decide what to do on failure, e.g. abort, restart the scan, restart the procedure, warn the user...

They should be able to be registered and unregistered within a scan or procedure.
 Migrated from GitLab issue https://gitlab.psi.ch/bec/bec/-/work_items/407

wakonig:
This PR migrates to the new beamline states. It builds the foundation for `Actors` (coming in a follow-up PR). A similar PR is in preparation for BEC Widgets to allow users to display the state of their beamline through small indicators

image

## CLI Interface 

Beamline states are composed of two components: A pydantic-based state config and the actual state class. For a user, it is sufficient to specify the pydantic model as the model references the state class. This design should also simplify the integration into BEC Widgets later on. 

closes #406


bec_lib/bec_lib/beamline_states.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        Configure the positioner condition.
+
+        Args:
+            device (str): The name of the positioner device.
+            min_limit (float): The minimum limit for the positioner.
+            max_limit (float): The maximum limit for the positioner.
+            tolerance (float): The tolerance for warning conditions (default is 0.1). When the positioner is within
+                               10% of the limits, a warning condition will be issued.
+            signal (str, optional): The name of the signal to monitor. If not provided, defaults to the device name.
```
wyzula-jan: why not to fallback first to hinted signal and then device name if hinted is not defined?

wakonig: done

bec_lib/bec_lib/bl_states.py
```diff
@@ -0,0 +1,338 @@
...
+        if val == "open":
+            return messages.BeamlineStateMessage(
+                name=self.name, status="valid", label="Shutter is open."
+            )
+        return messages.BeamlineStateMessage(
+            name=self.name, status="invalid", label="Shutter is closed."
+        )
+
+
+class DeviceWithinLimitsState(DeviceBeamlineState):
```
cappel89: I tried using the DeviceWithinLimitState, and checked the endpoint in REDIS. Somehow, I only saw invalid and warning updates. After some iterations, I realized that this is due to me setting tolerance to 1, which resulted in  a broad tolerance region, essentially no valid region. I am wondering if absolute tolerance might be more intuitiv.

bec_lib/bec_lib/bl_states.py
```diff
@@ -0,0 +1,338 @@
...
+    A state that checks if the shutter is open.
+
+    Example:
+        shutter_state = ShutterState(name="shutter_open")
+        shutter_state.configure(device="shutter1")
+        bec.beamline_states.add(shutter_state)
+    """
+
+    def evaluate(self, msg: messages.DeviceMessage, **kwargs) -> messages.BeamlineStateMessage:
+        val = msg.signals.get(self.signal, {}).get("value", "").lower()
```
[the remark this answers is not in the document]
wakonig: depending on how the device is integrated, you may need a custom beamline state. I don't think we should try and generalize the beamline state if the underlying devices are not generalized. The current implementation is a demo implementation for a shutter device. We have to see what the device reports in real operation and adjust it accordingly but I would be in favor of implementing the shutter devices consistently, rather than trying to patch it here. 

Regarding the second part of your question, yes, the signal selection can be removed.

bec_lib/bec_lib/bl_states.py
```diff
@@ -0,0 +1,338 @@
...
+    def _update_device_state(msg_obj: MessageObject, parent: DeviceBeamlineState) -> None:
+
+        # Since this is called from the Redis connector, we
+        assert parent.connector is not None
+
+        msg: messages.DeviceMessage = msg_obj.value  # type: ignore ; we know it's a DeviceMessage
+        out = parent.evaluate(msg)
+        if out is not None and out != parent._last_state:
+            parent._last_state = out
+            parent.connector.xadd(
```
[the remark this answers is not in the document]
wakonig: this is expected. Redis does not strictly enforce the message length and may decide to keep a few more messages

bec_server/bec_server/file_writer/file_writer_manager.py
```diff
@@ -134,6 +145,61 @@ def _device_configuration_callback(msg, *, parent: FileWriterManager):
...
+                    parent=self,
+                    from_start=True,
+                )
+                self.beamline_state_subscriptions.add(state.name)
+        # Unregister from states that are no longer needed
+        for topic in current_subscriptions - needed_subscriptions:
+            self.connector.unregister(
+                MessageEndpoints.beamline_state(topic), cb=self._beamline_state_callback
+            )
+            self.beamline_state_subscriptions.remove(topic)
```
cappel89: I tested removing a beamline state, and it was still in the H5 file. Simply adding 

```
self.beamline_states.pop(topic, None)
```
as a next line solves it. I tested it.

wakonig: good catch! - fixed it and extended the tests

PR #697, opened 2025-12-09
Feature/pretty print device

wakonig:
This PR improves the device representation in IPython:

| Old | New |
|--------|--------|
| image | image |

It also separates the pretty print from normal access and access within containers. In the past, this has caused quite some problems with lists as a simple `.get_devices_with_tags` call can demonstrate:

| Old | New |
|--------|--------|
| image | image |

For devices with large numpy arrays, the representation has improved as well:

| Old | New |
|--------|--------|
| image | image |

### DIFF
--- a/bec_lib/bec_lib/device.py
+++ b/bec_lib/bec_lib/device.py
@@ -11,6 +11,7 @@ import threading
 import time
 import uuid
 from collections import defaultdict, namedtuple
+from datetime import datetime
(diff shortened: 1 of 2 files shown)


wakonig:
> Do we want to also then include some pretty print on signals?

Probably not, at least not on a `.read()`. Otherwise people may not understand that it is a dictionary return value.

wyzula-jan (approved):
looks good 🚀🚀🚀

PR #701, opened 2025-12-11
Feat: Add Procedures HLI

### LINKED ISSUE #681: [FEAT]: Add high-level interface for launching procedures
As a user, I would like to have an easy way of launching an already defined macro as a new procedure. The macro may need to receive input arguments. I want to stay informed about the current status of the procedure execution, potentially through a status object, similar to how it is implemented for scans. 

A potential API could look something like this 

```python
status = bec.proc.run_macro("<macro_name>", arg1, kwarg1=kwarg1) 
status.wait()

assert status.done is True

```

The run_macro interface would check if the script is available and upload it for the execution. This ensures that the procedure worker would also be able to see local scripts, outside of the beamline repo.

d-perl:
closes #681 

To test procedures locally you can start the BEC server with: `bec-server start --use-in-process-proc-worker`, this will avoid the use of podman.

Introduces a high level interface in the BEC client to manage procedures, with a convenience function to run a loaded macro

Also starts the procedure runner as part of the scan server

You can operate it from `bec.proc`, e.g.:

```python
• default@bec [1/1] ❯❯ st = bec.proc.request_new("sleep", ((),{"time_s":5}))

• default@bec [2/1] ❯❯ st
Out[2]: ProcedureStatus for sleep: Scheduled

• default@bec [3/1] ❯❯ st
Out[3]: ProcedureStatus for sleep: Running

• default@bec [4/1] ❯❯ st
Out[4]: ProcedureStatus for sleep: Finished

• default@bec [5/1] ❯❯ st = bec.proc.request_new("doesn't exist")

• default@bec [6/1] ❯❯ st
Out[6]:
ProcedureStatus for doesn't exist: Rejected
ERROR:
Procedure doesn't exist not known to the server. Available: ['log execution message args', 'run scan', 'sleep', 'run_script', 'run_macro']

• default@bec [7/1] ❯❯ st = bec.proc.run_macro("macro_test", 2, 3)
```

and get a list of available procedures like:

```
• default@bec [1/1] ❯❯ bec.proc.available_procedures()

Available procedures and their signatures:
------------------------------------------

'log execution message args':
    (*args, **kwargs)

'run scan':
    (scan_name: str, args: tuple, parameters: dict)

'sleep':
    (time_s: float)

'run_script':
    (script_id: str)

'run_macro':
    (macro_name: str, params: tuple[tuple, dict] | None = None)
```

### DIFF
--- a/bec_lib/bec_lib/client.py
+++ b/bec_lib/bec_lib/client.py
@@ -141,7 +141,6 @@ class BECClient(BECService):
         self._init_procedure_hli: bool = self.__init_params[
             "config"
         ]._config_model.procedures.enable_procedures
-        print(self._init_procedure_hli)
         self._name = name
         self.device_manager = None
         self.queue = None
--- a/bec_lib/bec_lib/macro_update_handler.py
+++ b/bec_lib/bec_lib/macro_update_handler.py
@@ -89,12 +89,6 @@ def has_executable_code(content: str) -> tuple[bool, int | None]:
     return False, None
 
 
-class _Macro(TypedDict):
-    cls: Callable
-    source: str
-    fname: str
-
-
 class MacroUpdateHandler:
     def __init__(self, macros: UserMacros):
         """
--- a/bec_lib/bec_lib/procedures/helper.py
+++ b/bec_lib/bec_lib/procedures/helper.py
@@ -58,7 +58,7 @@ class ProcedureStatus:
         if self.state not in [
             ProcedureState.REQUESTED,
             ProcedureState.RUNNING,
-            ProcedureState.RUNNING,
+            ProcedureState.SCHEDULED,
         ]:
             raise ValueError(f"A procedure which is already {self.state} cannot be cancelled!")
         self._helper.request.abort_execution(self._id)
@@ -151,6 +151,7 @@ class _HelperBase:
                 self._callback_ids[msg_.execution_id].state = ProcedureState.RUNNING
             elif msg_.action == "Aborted":
                 self._callback_ids[msg_.execution_id].state = ProcedureState.FAILED
+                self._callback_ids[msg_.execution_id].set_error("Aborted by user.")
                 del self._callback_ids[msg_.execution_id]
             elif msg_.action == "Finished":
                 if msg_.error is not None:
--- a/bec_lib/bec_lib/messages.py
+++ b/bec_lib/bec_lib/messages.py
@@ -1358,12 +1358,12 @@ class ProcedureClearUnhandledMessage(ProcedureAbortMessage):
 
     One and only one of the args should be supplied.
         Args:
-            queue (str | None): the procedure queue to abort
-            execution_id (str | None): the procedure execution to abort
-            abort_all (bool | None): abort all procedures if true
+            queue (str | None): the unhandled procedure queue to clear
+            execution_id (str | None): the unhandled procedure queue to clear
+            abort_all (bool | None): clear all procedures if true
     """
 
-    ...
+    msg_type: ClassVar[str] = "procedure_clear_unhandled_message"
 
 
 class ProcedureWorkerStatusMessage(BECMessage):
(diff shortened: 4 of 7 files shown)


bec_lib/bec_lib/procedures/helper.py
```diff
@@ -9,22 +12,170 @@
...
+        self._helper = FrontendProcedureHelper(self._conn, monitor_responses=False)
+        self._state: ProcedureState = ProcedureState.REQUESTED
+        self._done: Event = Event()
+        self._error: str | None = None
+
+    @property
+    def done(self):
+        return self._done.is_set()
+
+    def wait(self, timeout_s: float | None = None):
```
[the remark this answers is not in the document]
d-perl: done

bec_lib/bec_lib/procedures/helper.py
```diff
@@ -9,22 +12,183 @@
...
 from bec_lib.messages import ProcedureQNotifMessage as QNotifMsg
 from bec_lib.messages import ProcedureRequestMessage as ReqMsg
+from bec_lib.messages import ProcedureStatusUpdate as StatUpd
+from bec_lib.messages import RequestResponseMessage as RespMsg
 from bec_lib.redis_connector import RedisConnector
 
 logger = bec_logger.logger
 
 
+class ProcedureState(Enum):
```
cappel89: What's the state if it's aborted? FAILED?

bec_lib/bec_lib/client.py
```diff
@@ -124,12 +132,16 @@ def __init__(
...
+        self.__init_params: _InitParams = {
             "config": config if config is not None else ServiceConfig(config_name="client"),
             "connector_cls": connector_cls if connector_cls is not None else RedisConnector,
             "wait_for_server": wait_for_server,
             "prompt_for_acl": prompt_for_acl,
         }
+        self._init_procedure_hli: bool = self.__init_params[
+            "config"
+        ]._config_model.procedures.enable_procedures
+        print(self._init_procedure_hli)
```
cappel89: please remove print

bec_lib/bec_lib/procedures/helper.py
```diff
@@ -9,22 +12,183 @@
...
 class _Request(_HelperBase):
+
+    def __init__(self, conn: RedisConnector, monitor_responses: bool = True) -> None:
+        super().__init__(conn, monitor_responses)
+
     def _xadd(self, ep: EndpointInfo, msg: BECMessage):
         self._conn.xadd(ep, msg.model_dump())
 
-    def procedure(self, msg: ReqMsg):
+    def procedure(
```
cappel89: What is the difference between the AbrtMsg and ClrMsg.. 
I looked at the messages itself, and it seems that they are the same..
Do we need both?

bec_server/bec_server/procedures/constants.py
```diff
@@ -65,7 +65,7 @@ class _CONTAINER:
 class _PROCEDURE:
     WORKER = _WORKER()
     CONTAINER = _CONTAINER()
-    MANAGER_SHUTDOWN_TIMEOUT_S = 2
+    MANAGER_SHUTDOWN_TIMEOUT_S: float | None = None
```
cappel89: What does this change imply?

bec_lib/bec_lib/macro_update_handler.py
```diff
@@ -89,6 +89,12 @@ def has_executable_code(content: str) -> tuple[bool, int | None]:
     return False, None
 
 
+class _Macro(TypedDict):
```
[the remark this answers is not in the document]
d-perl: It was used before https://github.com/bec-project/bec/pull/712 but wasn't cleaned up after rebasing. Removed it.

bec_lib/bec_lib/procedures/helper.py
d-perl: They carry the same information but go to separate endpoints and are used for different purposes. I updated the `msg_type` for that one though since it should be different.

cappel89 (approved):

PR #703, opened 2025-12-11
fix(atlas): use proper endpoints to avoid deprecation warnings

wakonig:
Moving hard-coded topic strings to proper endpoints


cappel89 (approved):
lg

PR #707, opened 2025-12-15
fix(dap): lmfit1d can pass any parameters with kwargs

wyzula-jan:
Adaptation that LMFIT parameters kwargs can be passed for lmfit1d_service, required for https://github.com/bec-project/bec_widgets/pull/987 ; requested by cSAXS

One can now pass multiple models for more complex fitting such as multiple Gaussians, fit background etc. Check https://github.com/bec-project/bec_widgets/pull/987 where are examples and showcase of high level API in `waveform.py`

### DIFF
--- a/bec_server/bec_server/data_processing/lmfit1d_service.py
+++ b/bec_server/bec_server/data_processing/lmfit1d_service.py
@@ -51,6 +51,7 @@ class LmfitService1D(DAPServiceBase):
         self.device_y = None
         self.signal_y = None
         self.parameters = None
+        self.override_parameters = None
         self.model_definitions: list[ModelDefinition] = []
         self._parameter_override_names = []
         self.model = self._build_model(model)
@@ -262,6 +263,12 @@ class LmfitService1D(DAPServiceBase):
                 override_params.pop(name, None)
         return override_params
 
+    def _has_complete_parameter_overrides(self) -> bool:
+        if self.override_parameters is None:
+            return False
+        param_names = set(getattr(self.model, "param_names", []))
+        return bool(param_names) and param_names.issubset(self.override_parameters.keys())
+
     def _resolve_device_and_signal(
         self, device: DeviceBase | str | None, signal: DeviceBase | str | None
     ) -> tuple[DeviceBase | str | None, DeviceBase | str | None]:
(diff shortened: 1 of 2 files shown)


bec_server/bec_server/data_processing/lmfit1d_service.py
```diff
@@ -354,10 +403,31 @@ def process(self) -> tuple[dict, dict]:
         y = self.data["y"]
 
         # fit the data
+        model_name = self.model.__class__.__name__
         if self.parameters:
-            result = self.model.fit(y, x=x, params=self.parameters)
+            logger.info(
```
wakonig: these logs could be potentially very verbose. Maybe put them on debug?

wyzula-jan: changed to debug

bec_server/bec_server/data_processing/lmfit1d_service.py
```diff
@@ -41,14 +44,267 @@ def __init__(self, model: str, *args, continuous: bool = False, **kwargs):
...
+        return override_params
+
+    def _build_parameters_from_overrides(self, overrides: lmfit.Parameters) -> lmfit.Parameters:
+        full_params = self.model.make_params()
+        return self._apply_override_params(full_params, overrides)
+
+    def _prepare_fit_params(self, x: np.ndarray, y: np.ndarray) -> lmfit.Parameters:
+        if self.parameters is None:
+            return self._guess_parameters(x, y)
+        if self.override_params is not None and len(self.override_params) > 0:
```
cappel89: If a Parameter object is passed, we need to check if the value is +-inf (I believe this to be the default from lmfit if no value is passed; https://lmfit.github.io/lmfit-py/parameters.html). It only makes sense to guess the parameters using the guess function if there is not value set, otherwise it should be skipped. Note: Afterwards, you have to be careful to then not overwrite the value guess by value = +-inf.

cappel89 (commented):
I think that this needs still some refactoring to improve readability. I find it quite complex to follow the code and I think it could be implemented simpler. Besides that, the most relevant comment is linked to when a Parameter object is passed, and how overrides are handled. Please check code comment about 'value' and its default which can be +-inf.

bec_server/bec_server/data_processing/lmfit1d_service.py
```diff
@@ -41,14 +44,259 @@ def __init__(self, model: str, *args, continuous: bool = False, **kwargs):
...
+        if not callable(guess_fn):
+            return
+        try:
+            component_guess = guess_fn(y, x=x)
+            params.update(component_guess)
+        except Exception as guess_exc:
+            name = component_name or component.__class__.__name__
+            logger.debug(f"lmfit guess failed for component={name}: {guess_exc}")
+
+    def _log_guess(self, guessed_params: lmfit.Parameters) -> None:
```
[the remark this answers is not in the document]
wyzula-jan: it is very useful for adhoc debugging if the fit is going off, I would keep it, it is just debug level of logs anyway.

Issue #660, opened 2025-11-25
[FEAT]: Add device config update message


wakonig:
## Feature Summary

Update messages should be published to inform about the device that is currently being added to the config. The message should contain the 
* device name
* Update index 
* Total number of devices in this config update

Point 1 provides the information to users which device may take very long to connect. Points 2 and 3 would allow to add a progress bar. 

## Proposed Solution

Add a new message type and endpoint. Publish the info when adding the devices on the device server. 

## Priority

mid

PR #720, opened 2025-12-22
Feature/config update abort

### LINKED ISSUE #587: [FEAT]: Default timeout for signals
## Feature Summary

For Ophyd, timeouts of classes are controlled through `__default_connection_timeout`. Unfortunately, they propose to change this default not per class instance, but only for all classes. This means that we may be able to configure for all EpicsSignal, a desired timeout. This has to be applied before (!) the class is instantiated.. So we might like to add some additional properties here to our configuration, or at least allow this to be easily changed. I pasted below the specific method that every 'Device' and 'Signal' has and which is suggest through comments in the code to set the default timeout time..

 -> Signal https://github.com/bluesky/ophyd/blob/main/ophyd/signal.py#L1102
-> Device https://github.com/bluesky/ophyd/blob/main/ophyd/device.py#L1042

### LINKED ISSUE #639: [FEAT]: Ability to interrupt config upload
We should add the ability to interrupt config updates

### LINKED ISSUE #721: [BUG]: tmux launch raises during restart

The recent update to the new tmux interface introduced a "leaked" error message during restart:

```
envs/bec-312/lib/python3.12/site-packages/libtmux/neo.py", line 242, in fetch_obj
raise exc.TmuxObjectDoesNotExist(
libtmux.exc.TmuxObjectDoesNotExist: Could not find object
```

The error message can be safely ignored but we should suppress it to not confuse users.

wakonig:
This PR adds the option to cancel a config update request. The "cancel" action is submitted automatically if the user "ctrl-c"s the config update. 

While a device init cannot be aborted, cancelling the update will lead to an early return. With the recent changes for the device initialization, a single device init should not take much longer than a second, so I believe this is a good compromise without making the device initialization too complicated. 

closes #639 
closes #721

The PR also adds the preparation for client-side updates during the config update: A new message has been added that is emitted throughout the config load. 
related to #660 

Moreover, the device model has been modified to also provide a timeout option (`connectionTimeout`). By default it is set to 5s but any device can override it in the config if needed. The timeout is forwarded to the `wait_for_connection` method. 
closes #587 

In a first iteration I wanted to stay conservative with the timeout but if we see that most device can work with, say 2s or even 1s, I would be very keen on reducing it.

### DIFF
--- a/bec_lib/bec_lib/config_helper.py
+++ b/bec_lib/bec_lib/config_helper.py
@@ -538,6 +538,7 @@ class ConfigHelper:
             '# - readoutPriority: Determines how the device is read out during a scan. Possible values are ["monitored", "baseline", "async", "on_request", "continuous"].\n'
             "#\n"
             "# Optional fields (defaults shown):\n"
+            "# - connectionTimeout: Connection timeout in seconds. Default is 5.\n"
             '# - description: A string description of the device. Default is "" (empty string).\n'
             "# - deviceConfig: A dictionary of configuration parameters specific to the device class. Default is None.\n"
             "# - deviceTags: A list/set of tags associated with the device. Default is an empty list/set.\n"
@@ -560,6 +561,7 @@ class ConfigHelper:
             "#     - frontend\n"
             "#     - motor\n"
             "#   enabled: true\n"
+            "#   connectionTimeout: 20\n"
             "#   readoutPriority: baseline\n"
             "#   onFailure: retry\n"
             "#   readOnly: false\n"
--- a/bec_server/bec_server/device_server/devices/devicemanager.py
+++ b/bec_server/bec_server/device_server/devices/devicemanager.py
@@ -415,7 +415,9 @@ class DeviceManagerDS(DeviceManagerBase):
         pipe = self.connector.pipeline()
         self.reset_device_data(obj, pipe)
         # Try to connect to the device, needs wait_for_all to include lazy signals e.g. AD detectors
-        raised_exc = self.connect_device(obj, wait_for_all=True)
+        raised_exc = self.connect_device(
+            obj, wait_for_all=True, timeout=dev.get("connectionTimeout", 5)
+        )
         # Publish device info with connect = True if no exception was raised during connection
         # Otherwise publish with connect = False
         connect = True if raised_exc is None else False
--- a/bec_lib/bec_lib/atlas_models.py
+++ b/bec_lib/bec_lib/atlas_models.py
@@ -40,6 +40,7 @@ class _DeviceModelCore(BaseModel):
     deviceClass: str
     readoutPriority: Literal["monitored", "baseline", "async", "on_request", "continuous"]
     deviceConfig: dict | None = None
+    connectionTimeout: float = 5.0
     description: str = ""
     deviceTags: set[str] = set()
     onFailure: Literal["buffer", "retry", "raise"] = "retry"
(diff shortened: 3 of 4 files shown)


bec_server/bec_server/device_server/devices/config_update_handler.py
```diff
@@ -15,45 +17,123 @@
...
+        cancel_event = request_info["cancel_event"]
+        future = request_info["future"]
+        active_request_id = request_info["request_id"]
+        cancel_event.set()
+        logger.info(f"Cancellation requested for config request {active_request_id}")
 
-    def parse_config_request(self, msg: messages.DeviceConfigMessage) -> None:
+        # Wait for the task to actually stop
+        try:
+            concurrent.futures.wait([future])
```
cappel89: Just a thought... 

In case a device is falsely integrated with wait_for_connection blocking indefinitely. Should we add a timeout (~10-20s) to avoid us blocking forever?

cappel89: And add additional logs, indicators that a device blocks in wait_for_connection

bec_server/bec_server/device_server/devices/devicemanager.py
```diff
@@ -122,7 +123,7 @@ def _get_device_class(dev_type: str) -> type:
         """Get the device class from the device type"""
         return plugin_helper.get_plugin_class(dev_type, [opd, ophyd])
 
-    def _load_session(self, *_args, **_kwargs):
+    def _load_session(self, *_args, cancel_event: threading.Event | None = None, **_kwargs):
```
cappel89: Moving forward, I think it would be really nice to add here device_config_progress updates similar to device_progress or scan_progress. This will allow for a nicer user experience upon reloading new devices.

cappel89 (commented):
PR looks good to me. I tested it more extensively without issues. Once tests passed, I can give it another look. 

One small comment below. I would suggest to create a new issue from it.

bec_lib/bec_lib/config_helper.py
```diff
@@ -617,7 +617,11 @@ def send_config_request(
         if wait_for_response:
             timeout = timeout_s if timeout_s is not None else self.suggested_timeout_s(config)
             logger.info(f"Waiting for reply with timeout {timeout} s")
-            reply = self.wait_for_config_reply(request_id, timeout=timeout)
+            reply = self.wait_for_config_reply(
+                request_id, timeout=timeout, send_cancel_on_interrupt=(action != "cancel")
+            )
+            if action == "cancel":
+                raise DeviceConfigError("Config update was cancelled by user.")
```
cappel89: One last comment:

DeviceConfigError(
    "Configuration loading was interrupted by the user; the configuration was flushed."
)

bec_lib/bec_lib/messages.py
```diff
@@ -906,6 +906,27 @@ def check_config(cls, values):
...
+        metadata (dict, optional): Additional metadata.
+
+    """
+
+    msg_type: ClassVar[str] = "device_initialization_progress_message"
+    device: str
+    finished: bool
+    success: bool
+    index: int
+    total: int
```
cappel89: Nice, 
just a brief remark. The message reports on device progress (device, finished and success) and in addition about the total progress (index and total). 
I am missing a flag that the full config update finished. 
Maybe 
`all_finished: bool`

PR #722, opened 2025-12-23
feat(scans): add hexagonal scan

wakonig:
Adding a hexagonal scan. Could sometimes be more useful than a simple grid scan. 

image

### DIFF
--- a/bec_server/bec_server/scan_server/scans.py
+++ b/bec_server/bec_server/scan_server/scans.py
@@ -6,7 +6,7 @@ import threading
 import time
 import uuid
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, Literal
+from typing import Any, Literal
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


bec_server/bec_server/scan_server/scans.py
```diff
@@ -1107,6 +1156,114 @@ def _calculate_positions(self):
         )
 
 
+class HexagonalScan(ScanBase):
+    scan_name = "hexagonal_scan"
+    required_kwargs = ["step_motor1", "step_motor2", "relative"]
```
cappel89: I suppose this was missed, please drop `step_motor1` and `step_motor2` as they are anyways args. 

Otherwise the syntax for the scan has to be:
```python
scans.hexagonal_scan(dev.samx, -5, 5, step_motor1=0.5, motor2=dev.samy,start_motor2=-5, stop_motor2=5  ,step_motor2=0.5, exp_time=0.1, relative=True)
```
instead of
``` python
scans.hexagonal_scan(dev.samx, -5, 5, 0.5, dev.samy, -5, 5, 0.5, exp_time=0.1, relative=True)
```
which currently raises for me..
Image

cappel89 (commented):
small comment, otherwise it looks good.

bec_server/bec_server/scan_server/scans.py
wakonig: good catch! - Thank you. I fixed it now

cappel89 (approved):
Looks good! 

I made an issue regarding the usage of `steps` and `motor_step` as we now have both cases in our core scans. Rather low in priority and not relevant for this PR.
https://github.com/bec-project/bec/issues/723

PR #765, opened 2026-02-11
Add ophyd_labels to device_info and summary

cappel89:
cSAXS has a rather complex device 'EPS', which is about to be merged to csaxs_bec (https://gitea.psi.ch/bec/csaxs_bec/pulls/134). Here PVs indicating status values, temperatures, etc. are merged together. This information is currently spread across the various signals of the device, and encoded in either 'docs' or 'labels. To get a nice overview of the relevant information, I would like to extend the device serializer and summary method to also show _ophyd_labels passed as `labels:set[str]` to the ophyd components.

I was also considering adding the 'doc' field of sub-devices, but then dropped the idea again as this would require to unfold all nested sub-devices which could make the table printout rather confusing.

Screenshot 2026-02-11 at 21 08 38

### DIFF
--- a/bec_lib/bec_lib/device.py
+++ b/bec_lib/bec_lib/device.py
@@ -1073,7 +1073,7 @@ class Device(OphydInterfaceBase):
                     signal_info.get("kind_str"),
                     signal_info.get("describe", {}).get("source"),
                     signal_info.get("describe", {}).get("dtype"),
-                    ", ".join(label for label in sorted(signal_info.get("labels", set(("",))))),
+                    ", ".join(label for label in signal_info.get("labels", [])),
                     signal_info.get("doc"),
                 )
                 kind_added = True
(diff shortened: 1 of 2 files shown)


bec_lib/bec_lib/device.py
```diff
@@ -1072,13 +1073,14 @@ def summary(self):
                     signal_info.get("kind_str"),
                     signal_info.get("describe", {}).get("source"),
                     signal_info.get("describe", {}).get("dtype"),
+                    ", ".join(label for label in sorted(signal_info.get("labels", set(("",))))),
```
wakonig: Why sorted? And why BEC signal labels not sorted?

cappel89: Done, casted to sorted list of strings already in the device_serializer.

PR #769, opened 2026-02-21
feat: add user feedback and leaderboard messages

wakonig:
Adding feedback and leaderboard message types. 
The feedback message bypasses the normal messaging service infrastructure and is therefore typed differently.


d-perl (approved):
Haha, is the idea to hook up minesweeper to this leaderboard?

wakonig:
Exactly! 😂

PR #795, opened 2026-03-13
refactor/fix: stream subscriptions in RedisConnector

d-perl:
Refactors the Redis stream subscription logic in the RedisConnector
- unifies all "normal" stream subscriptions: those which should read from the start of the stream are migrated to the normal subscriptions after the first read attempt, by reading up to the place those streams have been read already in the normal subscriptions, if it exists
- clarifies the organisation by moving the subscription logic methods to a separate class
- fixes some logical mistakes:
  - it is an error to register the same stream with the same callback more than once
  - no longer blocks separately on `from_start` streams after the first try
  - some small errors in the tests
- Garbage collects any subscriptions with dead references on each iteration of the respective message collection loop

Looks like some small patch in BEC widgets will be required where duplicate subs were previously ignored - workflow against BW branch https://github.com/bec-project/bec/actions/runs/23142709927

closes https://github.com/bec-project/bec/issues/790


wakonig (commented):
What an improvement, thank you! Could you pls add a test to verify that the newly added logic of merging from_start streams with normal streams after the first read does not skip messages?

wakonig (approved):

PR #809, opened 2026-04-02
refactor: pretty print scan queue

d-perl:
splits scan queue pretty print to make it easier to extract output

### DIFF
--- a/bec_lib/tests/test_queue_items.py
+++ b/bec_lib/tests/test_queue_items.py
@@ -48,7 +48,7 @@ def test_queue_storage_describe_queue_empty(queue_storage: QueueStorage):
     """
     Test describing an empty queue should not fail.
     """
-    assert queue_storage.describe_queue() == []
+    assert queue_storage.describe_queue() == "No current scan queue!"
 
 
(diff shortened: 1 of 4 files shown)


bec_lib/tests/test_queue_items.py
```diff
@@ -58,8 +58,8 @@ def test_queue_storage_describe_queue_item(queue_storage: QueueStorage, scan_que
     with mock.patch.object(queue_storage.scan_manager.connector, "get") as mock_get_current_queue:
         mock_get_current_queue.return_value = scan_queue_status_message
         out = queue_storage.describe_queue()
-        assert len(out) == 1
-        assert "primary queue / RUNNING" in out[0]
+        assert len(out) == 408
```
wakonig: not sure the length check is still useful. Before, it was checking that we indeed have the entire queue length mapped but now I find it quite difficult to decipher from "408" ;) -- I would just remove the assertion

Same for the other test

d-perl: done

wakonig (approved):

PR #813, opened 2026-04-13
fix: don't use staticmethod for subscriptions

d-perl:
This should fix the issues in https://github.com/bec-project/bec_widgets/pull/1096


wakonig (approved):
lgtm

PR #814, opened 2026-04-13
fix: lazy import utilities

wyzula-jan:
## Description

This PR fixes `bec_lib.utils.import_utils.lazy_import_from()` and makes the `bec_lib.utils` package import path lightweight.

`lazy_import_from()` now correctly supports both a single string and a sequence of strings. The previous implementation treated plain strings as iterables of characters and returned a one-shot generator for multi-name imports. The new implementation returns a single proxy for one name, a tuple of proxies for multiple names, and memoizes resolved attributes with `init_once=True`.

`bec_lib.utils.__init__` now exposes public utility helpers lazily. This is needed because Python imports `bec_lib.utils.__init__` before `bec_lib.utils.import_utils`, so the previous eager imports of `scan_utils`, `rpc_utils`, and `threading_utils` made importing the lazy import helper unnecessarily heavy.

## Type of Change

- Fix `lazy_import_from()` string handling, multi-name return behavior, and proxy memoization.
- Make `bec_lib.utils` public exports lazy to avoid eager imports when importing `bec_lib.utils.import_utils`.
- Add regression tests for lazy import behavior and import boundaries.

### DIFF
--- a/bec_lib/bec_lib/utils/__init__.py
+++ b/bec_lib/bec_lib/utils/__init__.py
@@ -1,3 +1,26 @@
-from bec_lib.utils.rpc_utils import user_access
-from bec_lib.utils.scan_utils import scan_to_csv, scan_to_dict
-from bec_lib.utils.threading_utils import threadlocked
+from importlib import import_module
+from typing import Any
+
+_LAZY_EXPORTS = {
+    "lazy_import": ("bec_lib.utils.import_utils", "lazy_import"),
+    "lazy_import_from": ("bec_lib.utils.import_utils", "lazy_import_from"),
+    "scan_to_csv": ("bec_lib.utils.scan_utils", "scan_to_csv"),
+    "scan_to_dict": ("bec_lib.utils.scan_utils", "scan_to_dict"),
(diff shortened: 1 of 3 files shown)


wakonig:
>bec_lib.utils.__init__ now exposes public utility helpers lazily. This is needed because Python imports bec_lib.utils.__init__ before bec_lib.utils.import_utils, so the previous eager imports of scan_utils, rpc_utils, and threading_utils made importing the lazy import helper unnecessarily heavy.

It is not like I don't agree with the changes but just out of curiosity: What exactly made it heavy in these imports? I would have thought functools should be quite fast

wyzula-jan:
> > bec_lib.utils.**init** now exposes public utility helpers lazily. This is needed because Python imports bec_lib.utils.**init** before bec_lib.utils.import_utils, so the previous eager imports of scan_utils, rpc_utils, and threading_utils made importing the lazy import helper unnecessarily heavy.
> 
> It is not like I don't agree with the changes but just out of curiosity: What exactly made it heavy in these imports? I would have thought functools should be quite fast

import_utils previously triggered bec_lib.utils.__init__, which imported scan_utils, and scan_utils imports typeguard

PR #819, opened 2026-04-14
Fix/startup lazy gui and dap

wyzula-jan:
## Description

This PR reduces `bec --nogui` startup time by deferring two startup-heavy paths that are not required immediately for a non-GUI shell session.

The GUI client is now created lazily when BEC starts with `--nogui`. The public `gui` object is still available, but BEC Widgets is not imported until the user actually calls something like `gui.show()` or `gui.new()`.

DAP plugin initialization is also deferred. `bec.dap` remains available, but the expensive DAP plugin setup is performed only when DAP functionality is first accessed.

In local benchmarking, this reduced `bec --nogui` startup time by about `2x`:

```bash
===== bec startup --nogui =====
Cases: bec
BEC branches: main,fix/startup-lazy-gui-and-dap
BEC Widgets branches: main
Benchmark 1: main
  Time (mean ± σ):      2.622 s ±  0.152 s    [User: 2.188 s, System: 0.346 s]
  Range (min … max):    2.495 s …  2.939 s    8 runs

Benchmark 2: fix/startup-lazy-gui-and-dap
  Time (mean ± σ):      1.319 s ±  0.009 s    [User: 1.130 s, System: 0.150 s]
  Range (min … max):    1.305 s …  1.333 s    8 runs


Summary
  fix/startup-lazy-gui-and-dap ran
    1.99 ± 0.12 times faster than main
```

## Type of Change

- Defer BEC Widgets GUI client import/creation for `bec --nogui` while keeping `gui` available.
- Defer DAP plugin initialization until first `bec.dap` access.

## How to test

- Run the existing unit tests for `bec_lib` and `bec_ipython_client`.
- Start `bec --nogui` and verify the client starts normally.
- From a `bec --nogui` session, call `gui.new()` or `gui.show()` and verify that the GUI can still be started on demand.
- Access `bec.dap` functionality and verify that DAP plugins still initialize when needed.

Startup benchmark with `hyperfine`:

Run this on `main`, record the result, then checkout this branch and run the same command again (I have the whole hyperfine wrapper, but this is minimal command to test it locally):

```bash
HOOK=$(mktemp /tmp/bec-startup-hook.XXXXXX.py)

cat > "$HOOK" <<'PY'
import os
import signal

print("BEC_BENCH_READY", flush=True)
os.kill(os.getpid(), signal.SIGTERM)
PY

hyperfine --warmup 2 --runs 8 --ignore-failure \
  "bec --nogui --dont-wait-for-server --post-startup-file $HOOK"

rm -f "$HOOK"
```

The benchmark starts `bec --nogui`, waits until the post-startup hook is executed, prints a marker, and then terminates the process. The non-zero exit code is expected because the hook intentionally stops the interactive process.

### Main
startup_main

### Fixed
startup_fix

## Potential side effects

In `--nogui` sessions, `gui` is initially a lazy proxy instead of an eagerly constructed `BECGuiClient`. Normal usage such as `gui.show()` and `gui.new()` should continue to work, but code that checks the exact type of `gui` before first use may observe the proxy type.

If BEC Widgets is unavailable or broken, the error is now raised when GUI functionality is first used rather than during `bec --nogui` startup.

`bec.dap` is initially a lazy wrapper. Normal DAP access should initialize the real DAP plugins transparently, but strict type checks or introspection may observe the lazy wrapper before first use.

## Additional Comments

This PR intentionally targets non-GUI shell startup. It does not change the default GUI startup behavior, except that GUI-related imports are avoided when starting with `--nogui`.

### DIFF
--- a/bec_ipython_client/bec_ipython_client/bec_startup.py
+++ b/bec_ipython_client/bec_ipython_client/bec_startup.py
@@ -1,6 +1,5 @@
 import os
 import sys
-import threading
 
 import numpy as np  # not needed but always nice to have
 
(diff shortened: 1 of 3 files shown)


d-perl (approved):
lgtm

PR #823, opened 2026-04-15
feat(device): support wildcard '*' in device name expansion

wakonig:
`"sam*"` was working but `"*"` is not a valid regex and was thus treated as a device. This PR fixes it to resolve `"*"` to all devices

### DIFF
--- a/bec_lib/bec_lib/devicemanager.py
+++ b/bec_lib/bec_lib/devicemanager.py
@@ -7,8 +7,8 @@ from __future__ import annotations
 import collections
 import contextlib
 import copy
+import fnmatch
 import functools
-import re
 import threading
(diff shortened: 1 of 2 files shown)


d-perl:
I don't fully understand. Is this supposed to work for `"sam*"` to expand to e.g. `"samx", "samy", "samz"`? I don't see where it would do that.

cappel89 (dismissed):

wakonig:
@d-perl should be fixed now. 
* `sam.*` now resolves to `[samx, samy, samz]` 
* `sa.*z` should include e.g. `[samz, sattrz]`

d-perl (approved):
LGTM!

PR #828, opened 2026-04-22
Fix error messages for sub-device access of setters

cappel89:
Changing config attributes of sub-devices is not supported for RPC devices, i.e. read_only, readoutPriority, limits etc. Previously, the error message for this was rather obscure:

image

This PR fixes this. Error messages for access of relevant attributes of sub-devices has been improved.

image

### DIFF
--- a/bec_lib/bec_lib/device.py
+++ b/bec_lib/bec_lib/device.py
@@ -79,13 +79,20 @@ class RPCError(AlarmBase):
 class NotImplementedOnSubdeviceError(AlarmBase):
     """Exception raised when a method is not implemented for a subdevice."""
 
-    def __init__(self, device: str, sub_device: str, method: str) -> None:
+    def __init__(
+        self, device: str, sub_device: str, method: str, additional_info: str | None = None
+    ) -> None:
+        compact_error_message = f"Method '{method}' is not implemented for subdevice '{sub_device}' of device '{device}'."
+        error_message = compact_error_message
+        if additional_info:
+            error_message += f" Additional info: {additional_info}"
+
         alarm = messages.AlarmMessage(
             severity=Alarms.MAJOR,
             info=messages.ErrorInfo(
                 exception_type="NotImplemented",
-                error_message=f"Method '{method}' is not implemented for subdevice '{sub_device}' of device '{device}'.",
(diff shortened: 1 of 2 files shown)


bec_lib/bec_lib/device.py
```diff
@@ -772,19 +798,30 @@ def _update_config(self, update: dict) -> None:
             update (dict): The update dictionary.
 
         """
+        if self.name != self.root.name:
+            raise NotImplementedOnSubdeviceError(
+                device=self.root.name,
+                sub_device=self.dotted_name,
+                method="_update_config",
+                additional_info=f"Received update: {update}",
+            )
```
d-perl, line 813: I guess this check could be simplified to a method `_raise_if_not_root(self, method: str, additional: str): ...` just to save a few lines of repetition

d-perl (dismissed):
looks good to me, one comment in code which could save a couple of lines if you wanted, optional

bec_lib/bec_lib/device.py
cappel89: agreed, I'll do this quickly.

bec_lib/bec_lib/device.py
```diff
@@ -76,6 +76,28 @@ def __init__(
         super().__init__(alarm, Alarms.MAJOR, handled=False)
 
 
+class NotImplementedOnSubdeviceError(AlarmBase):
+    """Exception raised when a method is not implemented for a subdevice."""
+
+    def __init__(
+        self, device: str, sub_device: str, method: str, additional_info: str | None = None
+    ) -> None:
+        compact_error_message = f"Method '{method}' is not implemented for subdevice '{sub_device}' of device '{device}'."
```
wakonig: Maybe add "access the root device {} instead."

bec_lib/bec_lib/device.py
```diff
@@ -885,6 +950,10 @@ def user_parameter(self) -> dict:
     @typechecked
     def set_user_parameter(self, val: dict):
         """set the user parameter for this device"""
+        if self.name != self.root.name:
```
wakonig: why not 

```python
self.root is not self
```

wakonig: @cappel89 friendly reminder ;)

cappel89: done

PR #832, opened 2026-04-24
Beamline actors part 1

d-perl:
Introduces a simplified version of the procedure manager and a sub-process worker to run Actors.

### DIFF
--- a/bec_lib/bec_lib/actors.py
+++ b/bec_lib/bec_lib/actors.py
@@ -1,0 +1,52 @@
+"""Definitions and protocols for the classes in bec_server.actors and in plugins."""
+
+from dataclasses import dataclass
+from enum import StrEnum
+from typing import Any, Protocol, runtime_checkable
+
+from typing_extensions import TypeAliasType  # only for 3.11
+
+from bec_lib.client import BECClient
+
+
+class ConditionCombination(StrEnum):
+    Any = "Any"
+    All = "All"
+
+
+class ActorCondition(Protocol):
+    def __call__(self, client: BECClient) -> Any:
+        """A callable which returns True if the condition is met and False if it is not."""
+
+
+@dataclass
+class ActorConditionSet:
+    conditions: set[ActorCondition]
+    combination_mode: ConditionCombination
+
+    def __call__(self, client: BECClient) -> bool:
+        if self.combination_mode == ConditionCombination.Any:
+            return any(condition(client) for condition in self.conditions)
+        return all(condition(client) for condition in self.conditions)
+
+
+class ActorAction(Protocol):
+    def __call__(self, client: BECClient) -> None:
+        """A callable which an actor calls if it has met the associated condition"""
+
+
+ActorActionTable = TypeAliasType(
+    "ActorActionTable", dict[ActorConditionSet | ActorCondition, ActorAction]
+)
+
+
+@runtime_checkable
+class Actor(Protocol):
+    client: BECClient
+    action_table: ActorActionTable
+
+    def __init__(self, client: BECClient, exec_id: str):
+        """Create the actor instance"""
+
+    def run(self):
+        """The core logic loop for the actor"""
--- a/bec_lib/bec_lib/actors.py
+++ b/bec_lib/bec_lib/actors.py
@@ -2,7 +2,7 @@
 
 from dataclasses import dataclass
 from enum import StrEnum
-from typing import Any, Protocol, runtime_checkable
+from typing import Any, Protocol
 
 from typing_extensions import TypeAliasType  # only for 3.11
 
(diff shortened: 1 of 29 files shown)


bec_server/bec_server/actors/actor.py
```diff
@@ -0,0 +1,96 @@
...
+            logger.info(f"Connecting {self.__class__.__name__} to '{endpoint.endpoint}'")
+            client.connector.register(endpoint, cb=self.evaluate)
+
+    def default_monitor_endpoints(self) -> set[EndpointInfo]:
+        return set()
+
+    def evaluate(self, *_, **__):
+        logger.info(f"{self.__class__.__name__} triggered")
+        if (now := time.monotonic()) < self.last_evaluated + self.min_delay_s:
+            logger.info("too little time elapsed since last trigger")
```
wakonig: not sure if we want to log here...

bec_lib/bec_lib/actors.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class ActorCondition(Protocol):
+    def __call__(self, client: BECClient) -> Any:
+        """A callable which returns True if the condition is met and False if it is not."""
+
+
+@dataclass
+class ActorConditionSet:
+    conditions: set[ActorCondition]
```
wakonig, line 36: maybe add some doc strings here to explain what the actorconditionset is and where it is used

bec_lib/bec_lib/endpoints.py
```diff
@@ -1637,6 +1640,45 @@ def procedure_logs(queue: str):
             endpoint=endpoint, message_type=messages.RawMessage, message_op=MessageOp.STREAM
         )
 
+    # Actors
+
+    @staticmethod
+    def actor_start_request():
+        """Endpoint to request starting a given actor."""
+        endpoint = f"{EndpointType.USER.value}/actors/request"
```
wakonig, line 1648: why a stream and not SEND for starting? Could lead to some side effects on restarts if the stream is not trimmed

bec_lib/bec_lib/endpoints.py
```diff
@@ -1637,6 +1640,45 @@ def procedure_logs(queue: str):
...
+            endpoint=endpoint,
+            message_type=messages.ActorStartRequestMessage,
+            message_op=MessageOp.STREAM,
+        )
+
+    @staticmethod
+    def actor_stop_request():
+        """Endpoint to request stopping a given actor."""
+        endpoint = f"{EndpointType.USER.value}/actors/stop"
+        return EndpointInfo(
```
wakonig, line 1659: same as above

bec_server/bec_server/actors/actor.py
```diff
@@ -0,0 +1,96 @@
...
+        self.name = name
+        self.exec_id = exec_id
+        self.client.connector.register(MessageEndpoints.actor_stop(exec_id), cb=self.stop)
+
+    def push_status(self, st: ProcedureWorkerStatus):
+        exec = self.exec_id if st == ProcedureWorkerStatus.RUNNING else None
+        push_status(self.client.connector, self.name, st, exec)
+
+    def evaluate(self, *_, **__):
+        for condition, action in self.action_table.items():
```
wakonig, line 44: pls add a doc string here to explain the logic of evaluation within actors

bec_lib/bec_lib/messages.py
```diff
@@ -2019,3 +2022,30 @@ class AvailableBeamlineStatesMessage(BECMessage):
 
     msg_type: ClassVar[str] = "beamline_state_update_message"
     states: list[BeamlineStateConfig]
+
+
+class ActorStartRequestMessage(BECMessage):
```
wakonig, line 2027: let's add some description to these messages

bec_server/bec_server/actors/actor.py
```diff
@@ -0,0 +1,96 @@
...
+from bec_server.procedures.oop_worker_base import push_status
+
+logger = bec_logger.logger
+
+
+class ActorBase(ABC):
+    client: BECClient
+    action_table: ActorActionTable
+
+    def __init__(self, client: BECClient, name: str, exec_id: str):
```
wakonig, line 21: do you foresee to "pause" an actor?

bec_server/bec_server/actors/actor.py
```diff
@@ -0,0 +1,96 @@
...
+from bec_lib.logger import bec_logger
+from bec_lib.messages import ProcedureWorkerStatus
+from bec_server.procedures.oop_worker_base import push_status
+
+logger = bec_logger.logger
+
+
+class ActorBase(ABC):
+    client: BECClient
+    action_table: ActorActionTable
```
wakonig, line 19: Some description of what the action_table is or should be used for is missing

wakonig (commented):
lgtm, only some minor comments regarding missing doc strings. I think the interface is very powerful, we just have to make sure that we will still understand it in 8 weeks :D 
As discussed offline, we should add it to the docs asap

bec_server/bec_server/actors/actor.py
d-perl: probably not, I think it's cleaner to just stop and start them. if it becomes important to start them with preserved state or without the second or two overhead we can re-evaluate

wakonig (approved):
lgtm

PR #834, opened 2026-04-24
Fix/v4 required kwargs

wakonig:
This PR removes the `required_kwargs` class attribute and instead relies on the scan signature to enforce kwargs. The scan signatures and examples have been adjusted to reflect the change in order. 

It does not change the functionality but simplifies the system a bit. 

Note that the PR targets the v4 branch, not main. If we merge v4 to main first, I will change it to main.


d-perl (approved):
nice!

PR #863, opened 2026-05-01
perf: simplify bec client mock and remove unused test devices

d-perl:
Changes the existing `ClientMock` which initialises a bunch of stuff to `FancyClientMock`, and introduces a simpler `ClientMock` which doesn't, which can be used in most cases. Removes many unused devices from the test config.

### DIFF
--- a/bec_lib/bec_lib/tests/test_config.yaml
+++ b/bec_lib/bec_lib/tests/test_config.yaml
@@ -3,305 +3,209 @@
 ############################################################
 
 eiger:
-  readoutPriority: monitored
-  deviceClass: ophyd_devices.SimCamera
-  deviceConfig:
-    device_access: true
-  deviceTags:
-    - detector
-  enabled: true
-  readOnly: false
-  softwareTrigger: true
+    readoutPriority: monitored
+    deviceClass: ophyd_devices.SimCamera
+    deviceConfig:
+        device_access: true
+    deviceTags:
+        - detector
+    enabled: true
+    readOnly: false
+    softwareTrigger: true
 
 ############## Cameras and detectors end here ##############
 
 dyn_signals:
-  readoutPriority: baseline
-  deviceClass: ophyd_devices.sim.sim.SynDynamicComponents
-  deviceConfig:
-  enabled: true
-  readOnly: false
+    readoutPriority: baseline
+    deviceClass: ophyd_devices.sim.sim.SynDynamicComponents
+    deviceConfig:
+    enabled: true
+    readOnly: false
 
 pseudo_signal1:
-  deviceClass: ophyd_devices.ComputedSignal
-  deviceConfig:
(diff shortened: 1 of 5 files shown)


bec_lib/bec_lib/tests/utils.py
```diff
@@ -103,7 +105,15 @@ def fermat_scan(
         pass
 
 
-class ClientMock(BECClient):
+class FancyClientMock(BECClient):
+    started = False
```
wakonig: so is FancyClientMock just a client with ScansMock? If so, I'd rather call it something like "ClientWithScansMock" or so

bec_lib/bec_lib/tests/test_config.yaml
```diff
@@ -3,305 +3,209 @@
...
-  deviceTags:
-    - detector
-  enabled: true
-  readOnly: false
-  softwareTrigger: true
+    readoutPriority: monitored
+    deviceClass: ophyd_devices.SimCamera
+    deviceConfig:
+        device_access: true
+    deviceTags:
```
wakonig, line 10: out of curiosity, what's the penalty of adding all these devices? Could you do a CI test run with all devices and your new client mocks?

d-perl: sure

bec_lib/bec_lib/tests/utils.py
d-perl: `FancyClientMock` is identical to the old `ClientMock`, the new `ClientMock` doesn't run `_start_services` and puts a couple extra `MagicMocks` in for the alarm handler, connector, and `get_global_var`

bec_lib/bec_lib/tests/test_config.yaml
d-perl: https://github.com/bec-project/bec/actions/runs/25212565141

bec_lib/bec_lib/tests/utils.py
d-perl: open to other names if you have any suggestions, but I think `ClientWithScans` doesn't quite fit since they both have scans

bec_lib/bec_lib/tests/test_config.yaml
d-perl: looks quite substantial, at least half of the speedup I would guess, maybe most of it even

wakonig: okay, then we remove them but we should investigate further why it is so slow

d-perl: they are deepcopied and reloaded into the device manager for every test which uses the client, which is most of them. I remember doing some profiling a while ago and a significant percentage of some tests was spent just in that deepcopy.

wakonig (dismissed):

wakonig (approved):

PR #864, opened 2026-05-03
feat: add scan modifier

wakonig:
## Description

This PR adds proper support for scan modifiers, including scan argument modification and gui config modification. To this end, the scan doc string and the examples are now created dynamically through the defined scan arguments. This allows modifiers to freely add and remove arguments whilst keeping everything up to date. 

## Type of Change

- Add support for scan modifiers, gui_config and scan argument modifications
- Auto-generate doc strings + signature + examples based on the currently active modifiers (v4 only!)

## How to test

- Run unit tests
- Add a scan modifier to the plugin repo

## Additional Comments

I'm preparing a PR for the plugin repo to also run e2e tests against a modifier. Once we've merged this PR, we can merge the plugin repo and add the e2e test in here.

### DIFF
--- a/bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
+++ b/bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
@@ -1,0 +1,590 @@
+"""
+Helpers for applying scan signature overrides from scan modifier plugins.
+
+This module builds the user-facing scan signature published by the scan manager
+and applies the same override-derived defaults again in the scan assembler
+before a direct scan class is instantiated. This keeps the published signature,
+input validation, and runtime construction behavior aligned.
+"""
+
+from __future__ import annotations
+
+import inspect
+from functools import lru_cache
+from typing import TYPE_CHECKING, Annotated, Any, Type, get_args, get_origin, get_type_hints
+
+from bec_lib.plugin_helper import get_scan_modifier_plugin
+from bec_lib.scan_args import ScanArgument
+from bec_lib.signature_serializer import deserialize_dtype, dict_to_signature, signature_to_dict
+from bec_server.scan_server.scans.scan_base import ScanBase
+
+if TYPE_CHECKING:
+    from bec_server.scan_server.scans.scan_modifier import ScanModifier
+
+
+@lru_cache(maxsize=None)
+def get_scan_modifier() -> Type[ScanModifier] | None:
+    """
+    Load the available scan argument modifier from the plugin.
+
+    Returns:
+        ScanModifier: The scan modifier class
+    """
+    modifier = get_scan_modifier_plugin()
+    return modifier
+
+
+def _convert_annotation(annotation: object) -> Annotated[Any, ScanArgument] | None:
+    """
+    Normalize an annotation to ``Annotated[..., ScanArgument(...)]``.
+
+    The scan signature override API operates on annotations that always carry
+    ``ScanArgument`` metadata so modifier code can treat original and injected
+    arguments consistently. Plain annotations are wrapped in a default
+    ``ScanArgument`` instance, while existing ``ScanArgument`` metadata is
+    preserved.
+
+    This is mainly done to keep the interface for scan signature modifiers consistent.
+
+    Args:
+        annotation (object): Serialized or native annotation object.
+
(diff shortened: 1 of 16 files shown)


bec_server/bec_server/scan_server/scans/scan_base.py
```diff
@@ -164,6 +165,7 @@ def __init__(
         instruction_handler: InstructionHandler,
         request_inputs: dict,
         system_config: dict,
+        scan_modifier: Type[ScanModifier] | None = None,
```
cappel89, line 168: Is this mostly for testing purposes or do you see a use case where custom implementations of the ScanModifier are passed to a scan?

bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
```diff
@@ -0,0 +1,576 @@
...
+def _get_scan_raw_doc(scan_cls: Type[ScanBase]) -> str:
+    """Return the scan-specific raw docstring, preferring non-generic class docs."""
+    class_doc = inspect.getdoc(scan_cls)
+    base_class_doc = inspect.getdoc(ScanBase)
+    base_init_doc = inspect.getdoc(ScanBase.__init__)
+
+    if class_doc and class_doc not in {base_class_doc, base_init_doc}:
+        return class_doc
+
+    return inspect.getdoc(scan_cls.__init__) or class_doc or ""
```
cappel89, line 435: This means that if we write both docs (__init__ and class_docs), we should put the relevant info in the docs of __init__, right?

bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
```diff
@@ -0,0 +1,576 @@
...
+
+
+def scan_doc_with_modifiers(scan_cls: Type[ScanBase]) -> str:
```
cappel89, line 438: Since scan docs are now writte based on the description field of ScanArgument, I would propose to add additonal description to some of the DefaultArgType, for example: "frames_per_trigger" and "burst_at_each_point". Otherwise, it is difficult to understand the difference without knowing the source code of scans. Those two I thought of immediately, maybe there are more.

bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
```diff
@@ -0,0 +1,576 @@
...
+        value = defaults[name] if name in defaults else _example_literal(name, arguments[name])
+        rendered = repr(value) if not isinstance(value, str) else value
+        if parameter.kind is inspect.Parameter.KEYWORD_ONLY:
+            if name not in defaults:
+                minimum_parts.append(f"{name}={rendered}")
+            full_parts.append(f"{name}={rendered}")
+        else:
+            if name not in defaults:
+                minimum_parts.append(rendered)
+            full_parts.append(rendered)
```
cappel89: This logic seems to fail for _v4_acquire and _v4_timescan.

bec_server/bec_server/scan_server/scans/scan_base.py
wakonig: no, the scan modifier is always injected into the scan. The scan assembler fetches the modifier class and injects it. This is needed as the scan hook decorator needs to redirect its own scan lifecycle hooks to modified hooks

bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
wakonig: this is how it was in the past, yes. Now with the dynamic docs, there is no need to specify everything in init and it might be more useful to simply describe the scan as class doc. Either way is fine though

bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
wakonig: done

bec_server/bec_server/scan_server/scans/scan_argument_modifier.py
wakonig: should be fixed now

cappel89 (approved):
lg

PR #831, opened 2026-04-23
Add Beamline State Machine and AggregatedState

cappel89:
Draft for AggregatedStates and state machine. To test, use a config like the following one:

```yaml
alignment: # AggregatedStateConfig -> can have different labels and for each label, different devices
  transition_metadata: # optional field for metadata for each label
    field: value 
  devices:
    samx:
      value: 0
      abs_tol: 0.1
      low_limit:
        value: -20
        abs_tol: 0.1
      high_limit:
        value: 20
        abs_tol: 0.1
      signals:
        velocity:
          value: 5
          abs_tol: 0.1
    bpm4i:
      value: 100
      abs_tol: 10
measurement:
  devices:
    samx:
      value: 19
      abs_tol: 0.1
      signals:
        velocity:
          value: 20
    samy:
      value: 0
      abs_tol: 0.1
test:
  devices:
    samy:
      value: 0
      abs_tol: 0.1
    bpm4i:
      value: 100
      abs_tol: 10
```
Assuming this config is stored under `state_config.yaml`, you can now load it as follows
```python
filename = "./state_config.yaml"
bec.state_machine.load_from_config(name="my_state", config_path=filename)
```

PR #866, opened 2026-05-04
Add AggregatedStates and  state machine

cappel89:
# Summary

This PR introduces a new state: AggregatedStates, and a cli tool "state_manager" to allow users to load states from a config definition. 

Configs must either be passed as dicts or YAML configs with matching structure as written below.

## A few remarks

- The AggregatedState accepts SIgnal or Device specs. This can either be directly a 'value' or 'at' which maps to a user parameter on the device.  
- Extended the interface to print bl_states. `bec.beamline_states.show_all()` now shows a collapsed view, and `bec.<state_name>.describe()` gives a detailed view of a state.

image

Example of the YAML configuration file:
``` yaml
bl_transition_states:
  bl_state_class: AggregatedState
  config:
    evaluation_method: any
    states:
      alignment:
        devices:
          samx:
            value: 0
            abs_tol: 0.1
            signals:
              velocity:
                value: 5
                abs_tol: 0.1
          bpm4i:
            value: 100
            abs_tol: 10
          samy:
            at: in
            abs_tol: 0.1
samx_within_limits:
  bl_state_class: DeviceWithinLimitsState
  config:
    device: samx
    signal: samx
    low_limit: -1
    high_limit: 1
    tolerance: 0.1
```
Assuming this config is stored under `state_config.yaml`, you can now load it as follows
```python
filename = "./state_config.yaml"
bec.state_machine.load_from_config(config_path=filename)
```

#closes #831

## TO test

There is also a branch with a draft to extend the beamline_states widget. This allows for easier testing of the new state, and loading configs conveniently. 

image


bec_lib/bec_lib/bl_states.py
```diff
@@ -322,6 +381,366 @@ def _update_device_state(self, msg_obj: MessageObject) -> messages.BeamlineState
...
+        raise ValueError(
+            f"Invalid signal source '{source}', please use 'readback', 'configuration', or 'limits'."
+        )
+
+    def _get_device_manager(self):
+        if self.device_manager is None:
+            # pylint: disable=import-outside-toplevel
+            from bec_lib.client import BECClient
+
+            bec = BECClient()
```
d-perl: should this check if the client is started?

bec_server/bec_server/scan_server/scans/state_transition_scan.py
```diff
@@ -0,0 +1,265 @@
...
+        self.state_name = state_name
+        self.target_label = target_label
+        # Check if the state and the target label exists, if yes, fetch the configuration for the target state
+        self.config_for_label = self._fetch_config_for_label(state_name, target_label)
+
+        # We need to sort the devices and signals in the config, and identify which of them are motor setpoint/readback pairs
+        # and which of them are just readouts and thereby can not be set within the transition.
+        self._signals_to_set: list[Tuple[Signal, Any]] = []
+        self._limits_to_set: dict[str, Tuple[Positioner, float, float]] = {}
+        self._devices_to_set: list[Tuple[Positioner, float]] = []
```
d-perl, line 82: are these definitely `Positioner`s? The logic later suggests they can be any device with `set()`?

bec_lib/bec_lib/bl_states.py
```diff
@@ -322,6 +381,366 @@ def _update_device_state(self, msg_obj: MessageObject) -> messages.BeamlineState
         return self.evaluate(msg)
 
 
+SignalSource = TypeVar("SignalSource", bound=Literal["readback", "configuration", "limits"])
```
d-perl: I don't think this is meant to be a `TypeVar`, as far as I can see it isn't used in parametrisation anywhere? I think it can just be `SignalSource = Literal["readback", "configuration", "limits"]`

otherwise the implication is that a given instance of `AggregatedState` can only have one kind of signal source for the entire state, and it should be `class AggregatedState(BeamlineState[AggregatedStateConfig], Generic[SignalSource])`, but I don't think that's what you meant, it seems more like just an enum or literal

bec_lib/bec_lib/bl_states.py
```diff
@@ -322,6 +381,366 @@ def _update_device_state(self, msg_obj: MessageObject) -> messages.BeamlineState
...
+        try:
+            msg = None
+            self._build_rules()
+            affected_labels = self._fill_cache()
+            msg = self.evaluate(affected_labels=affected_labels)
+        except Exception as exc:
+            self._handle_state_exception(exc)
+
+        if msg is not None:
+            self._emit_state(msg)
```
d-perl, line 1036: `msg = None` should be outside the `try`

bec_lib/bec_lib/bl_states.py
```diff
@@ -322,6 +381,366 @@ def _update_device_state(self, msg_obj: MessageObject) -> messages.BeamlineState
...
+        super().stop()
+
+    def _update_aggregated_state(
+        self, msg_obj: MessageObject, device: str, source: SignalSource, **_kwargs
+    ) -> None:
+        try:
+            msg: messages.DeviceMessage = msg_obj.value  # type: ignore ; we know it's a DeviceMessage
+            affected_labels = self._cache_message(device, source, msg)
+            if affected_labels:
+                msg = self.evaluate(affected_labels=affected_labels)
```
d-perl: now it's not a DeviceMessage anymore, it should get a new name

bec_lib/bec_lib/bl_states.py
```diff
@@ -322,6 +381,366 @@ def _update_device_state(self, msg_obj: MessageObject) -> messages.BeamlineState
...
+            return None
+        # We need to always extend the affected labels with the current labels,
+        # as the signal that updated might be not relevant for the currently active state,
+        # but the state should still be checked for validity.
+        affected_labels.update(self._current_labels)
+        matching_labels = [label for label in affected_labels if self._label_matches(label)]
+        if matching_labels:
+            self._current_labels = matching_labels
+            state_msg = messages.BeamlineStateMessage(
+                name=self.config.name, status="valid", label="|".join(matching_labels)
```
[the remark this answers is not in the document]
d-perl: not sure if copilot is right or not here, but it is probably good to sort them for display anyway

bec_lib/bec_lib/bl_states.py
d-perl, line 95: I know this isn't in this PR but I think these annotations are wrong, it should be `str|None` since it is always resolved by the validator

d-perl (commented):
I think the concept is good, some comments in code

bec_server/bec_server/scan_server/scans/state_transition_scan.py
```diff
@@ -0,0 +1,342 @@
...
+        self.state_name = state_name
+        self.target_label = target_label
+
+        # We need to sort the devices and signals in the config, and identify which of them are motor setpoint/readback pairs
+        # and which of them are just readouts and thereby can not be set within the transition.
+        self._signals_to_set: list[Tuple[Signal, Any]] = []
+        self._limits_to_set: dict[str, Tuple[Positioner, float, float]] = {}
+        self._devices_to_set: list[Tuple[Positioner, float]] = []
+
+    # pylint: disable=protected-access
```
wakonig: I don't see any protected access outside of the class

bec_lib/bec_lib/client.py
```diff
@@ -244,6 +246,7 @@ def _start_services(self):
         self.dap = DAPPlugins(self)
         self._update_username()
         self.beamline_states = BeamlineStateManager(client=self)
+        self.state_machine = BeamlineStateMachine(manager=self.beamline_states)
```
wakonig: I'm wondering if we should pack it under beamline states. What do you think?

cappel89: done!

PR #873, opened 2026-05-05
feat: implement AtlasActiveClientEmitter for tracking active clients …

wakonig:
## Description

* Add an emitter for active clients metric in BEC
* Emit the currently active account
* Emit connection status for BEC Atlas


cappel89 (approved):
Looks good, I suppose we should see metrics updates after redeployments.

PR #877, opened 2026-05-08
builtin actor manager and scan interlock actor

d-perl:
Adds a `BuiltinActorManager` in which we can run permanently-running hardcoded actors.
Adds an implementation of such, a base BeamlineStateActor and a ScanInterlockActor which locks the primary queue when specified beamline states go out of range.

### DIFF
--- a/bec_ipython_client/tests/end-2-end/test_actors_e2e.py
+++ b/bec_ipython_client/tests/end-2-end/test_actors_e2e.py
@@ -1,0 +1,100 @@
+from __future__ import annotations
+
+import time
+from typing import TYPE_CHECKING
+
+import pytest
+
+from bec_lib.bl_states import DeviceWithinLimitsStateConfig
+from bec_lib.builtin_actor_hli import BuiltinActorHli
+from bec_lib.logger import bec_logger
+from bec_lib.messages import ScanQueueStatus, ScanQueueStatusMessage
+
+logger = bec_logger.logger
+
+if TYPE_CHECKING:  # pragma: no cover
+    from ophyd_devices.sim.sim_test_devices import SimDeviceWithSignalDelay
+
+    from bec_ipython_client.main import BECIPythonClient
+
+
+@pytest.fixture
+def bec_with_delay_device(bec_ipython_client_fixture):
+    bec = bec_ipython_client_fixture
+    bec.builtin_actors.scan_interlock.enabled = True
+    dev = bec.device_manager.devices
+    config = {
+        "ramp_up": {
+            "deviceClass": "ophyd_devices.sim.sim_test_devices.SimDeviceWithSignalDelay",
+            "deviceConfig": {},
+            "readoutPriority": "baseline",
+            "deviceTags": {"signal delay", "ramp up"},
+            "enabled": True,
+            "readOnly": False,
+        }
+    }
+    bec.device_manager.config_helper.send_config_request(action="add", config=config)
+    dev.ramp_up.min_val.put(0)
+    dev.ramp_up.max_val.put(400)
+    dev.ramp_up.value.put(400)
+    dev.ramp_up.delay.put(1)
+    dev.ramp_up.rampup_time.put(3)
+    yield bec, dev.ramp_up
+    dev.ramp_up.stop()
+
+
+@pytest.fixture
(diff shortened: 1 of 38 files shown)


bec_ipython_client/tests/end-2-end/test_actors_e2e.py
```diff
@@ -0,0 +1,100 @@
...
+        if pred():
+            return True
+        time.sleep(timeout / retries)
+    return pred()
+
+
+# pylint: disable=protected-accesstest
+@pytest.mark.timeout(100)
+def test_scan_interlock(
+    ramp_up_bl_state: tuple[BECIPythonClient, SimDeviceWithSignalDelay], bec_with_delay_device
```
wakonig, line 58: can we extend it to run an entire scan? Would be helpful to check whether our restart logic works

bec_server/bec_server/actors/scan_interlock.py
```diff
@@ -0,0 +1,79 @@
...
+            self._update_cache()
+            self._update_watched_states_in_redis()
+        super(BlStateActor, self).evaluate()
+
+    def some_mismatch_action(self, client: BECClient):
+        if self.client.queue is None or self.lock_id is not None:
+            return
+        self.lock_id = str(uuid4())
+        self.client.queue.add_queue_lock(
+            queue="primary", reason="ScanInterlockActor", lock_id=self.lock_id
```
wakonig: the reason should rather be the name of the failed state, e.g. shutter is closed and lock id should be ScanInterlockActor

bec_server/bec_server/actors/scan_interlock.py
```diff
@@ -0,0 +1,79 @@
...
+class ScanInterlockActor(BlStateActor):
+    """Sets a scan lock on the primary queue if any of the the states in the state_table don't match
+    the required value. Removes the lock if all of them match."""
+
+    def __init__(self, client: BECClient, name: str, exec_id: str):
+        self.state_table = {}
+        super().__init__(client, name, exec_id)
+        self.lock_id: str | None = None
+        self._update_watched_states_in_redis()
+        self.client.connector.register(
```
wakonig: when I restart the server, all states are flushed. This can be quite dangerous. I think it would be better if we could recover the last state config

bec_lib/bec_lib/bl_state_manager.py
```diff
@@ -200,6 +200,16 @@ def delete(self, state_name: str) -> None:
             self._delete_state(state_name)
             self._publish_states()
 
+    def get_status_by_name(self, name: str) -> messages.BlStateStatus | None:
```
wakonig, line 203: unrelated to the PR but maybe we can add it here as well: It would be nice to check if a state is already registered, something like `state_exists(state_name: str)` or so. Alternatively, a flag in `add` to skip if it exists already to avoid updating an existing config

d-perl: this function already provides that by returning none if it doesn't exist

bec_server/bec_server/actors/builtin_actor_manager.py
```diff
@@ -0,0 +1,71 @@
...
+logger = bec_logger.logger
+
+
+class BuiltinActorManager:
+    """A simple manager for builtin actors which are always available - only handles enabling and
+    disabling"""
+
+    def __init__(self, bootstrap_server: str) -> None:
+        host, port = bootstrap_server.split(":")
+        self._client = BECClient(
```
wakonig, line 41: builtin actors are running on the same process, no? In that case, we can just use the singleton client without init or directly inject the client

wakonig (commented):
The enabling / disabling logic seems to be broken: I tried to play around with the flags in a second client but it seems not work as expected. 

Scenario I: I had the interlock enabled and triggered, a scan was pending. I disabled the interlock but the scan did not continue until the returned to a valid state. 

Scenario II: I started with a disabled interlock actor and enabled it mid scan. I would expect that the scan is stopped and repeated but it simply continued without respecting the beamline state.

wakonig (dismissed):
works as expected! 

As discussed offline, there are still some open points but I would suggest to merge it as is now and refine it in another PR to allow the widget development to proceed in the meantime.

wakonig (approved):

PR #885, opened 2026-05-12
Bugfix for set not raising during a mv command.

cappel89:
PR to fix a bug that the `mv` command would not properly raise if `.set` failed. The `umv` command was handling this properly as raising the exception was properly captured and awaited for in contrast to the `mv` which occurs fully asynchronously.


wakonig (dismissed):
lgtm

wakonig (approved):

PR #887, opened 2026-05-12
fix: clean procedure worker shutdown

d-perl:
Makes sure that the procedure worker's `RedisConnector` is shut down when the worker is aborted. Mainly comes up in tests; I couldn't reproduce it manually, but I think it happens when pytest intercepts cleanup and the test finishes before the worker context manager exits, which is what would normally call `_conn.shutdown()`.

Additionally, Podman doesn't seem to respect `--replace` option if there is some error state from a previous cleanup, resulting in flaky tests. This change forces the cleanup before starting a container but doesn't resolve the root issue.

Also adds names to the RedisConnector threads to make it easier to identify which is which.

### DIFF
--- a/bec_server/bec_server/procedures/worker_base.py
+++ b/bec_server/bec_server/procedures/worker_base.py
@@ -47,7 +47,7 @@ class ProcedureWorker(ABC):
         self._active_procs_endpoint = MessageEndpoints.active_procedure_executions()
         self.status = ProcedureWorkerStatus.IDLE
         self._redis_server = server
-        self._conn = RedisConnector([server])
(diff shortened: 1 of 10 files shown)


bec_server/bec_server/procedures/worker_base.py
```diff
@@ -86,6 +86,7 @@ def logs(self) -> list[str]:
     def abort(self):
         """Abort the entire worker"""
         self._aborted.set()
+        self._conn.shutdown()
```
d-perl, line 89: this is the crucial line

wakonig (approved):
lgtm

PR #889, opened 2026-05-13
Adjust progressbar time resolution depending on elapsed time

d-perl:
Change the display format for elapsed time in the progressbar depending on the value, to display millis if the time is low


wakonig (approved):
nice!!

PR #891, opened 2026-05-15
fix: add optional separator to dynamic metrics

d-perl:
Add a string (default `"-"`) to the beginning of dynamic metric names, so that they are separated when concatenated with the group name in the ingestor.

### DIFF
--- a/bec_lib/bec_lib/messages.py
+++ b/bec_lib/bec_lib/messages.py
@@ -1225,11 +1225,13 @@ class DynamicMetricMessage(BECMessage):
     timestamp: float = Field(default_factory=time.time)
 
     @classmethod
-    def from_dict(cls, metrics: dict[str, str | int | float | bool]):
+    def from_dict(cls, metrics: dict[str, str | int | float | bool], separator: str = "-"):
         return cls.model_validate(
             {
                 "metrics": {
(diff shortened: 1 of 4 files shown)


bec_lib/bec_lib/messages.py
```diff
@@ -1225,11 +1225,13 @@ class DynamicMetricMessage(BECMessage):
...
 
     @classmethod
-    def from_dict(cls, metrics: dict[str, str | int | float | bool]):
+    def from_dict(cls, metrics: dict[str, str | int | float | bool], separator: str = "-"):
         return cls.model_validate(
             {
                 "metrics": {
-                    k: {"value": v, "type_name": type(v).__name__} for k, v in metrics.items()
+                    # Metric names are concatenated with the group in the ingestor
+                    separator + k: {"value": v, "type_name": type(v).__name__}
```
wakonig, line 1233: maybe something like

```python 
separator + k.removeprefix(separator)
```

To ensure that it is not added twice?

d-perl: do you think that's necessary? if someone just follows another use of `RedisConnector.publish_metrics` from the code they will not add it twice, it feels a bit overcomplicated

wakonig (approved):

PR #893, opened 2026-05-16
Feature/scan bundler device progress

wakonig:
## Description

This PR merges the scan progress update for monitored readouts and device progress into a single update for "scan progress". The scan bundler now automatically subscribes to the device for the scan and forwards the update accordingly. 

## How to test

- Run unit tests
- Run a scan with device progress and one with normal monitored progress. Check the plugin repo for a scan with device progress.

## Additional Comments

Once merged, we can simplify the subscription logic in BW.


wyzula-jan (approved):
works with https://github.com/bec-project/bec_widgets/pull/1178

PR #894, opened 2026-05-18
fix(ipython_live_updates): accept None as valid states and return early

wakonig:
## Description

The ipython live update was asserting that states that were checked previously are still not None. However, None is a valid state and depending on the OS schedule may go back to None. This PR accepts None as valid state and simply returns early.


d-perl (approved):
Ah...

PR #907, opened 2026-05-21
feat: add logger import and initialize logger in v4_scan template

wakonig:
## Description

Adding the logger to the scan template.


cappel89 (approved):
lg

PR #914, opened 2026-05-26
feat(signal): add support for sending messages directly to phone numbers

wakonig:
## Description

This PR adds support for sending messages through the signal service merely by providing phone numbers, that is without setting up signal groups. This should allow us to receive updates during debugging sessions. 

## Type of Change

- Allow empty scopes for Signal
- Add phone number parsing to signal message objects

## How to test

- Run unit tests
- Connect to atlas dev and send messages to your phone number

## Definition of Done
- [ ] Documentation is up-to-date. <- I will add a new PR to bec_docs later


cappel89 (approved):
Looks good. So typos with spaces or invalid characters also pass. Only that last one failed, but I think this is a good feature IMO. Only the last one failed from the screenshot below.

PR #919, opened 2026-05-27
feat: implement dynamic string metrics

d-perl:
- allows users to supply the possible values for string metrics so that they can be properly processed
- example implementation for atlas host using an enum for possible options and convenience


wakonig (approved):
awesome :D

PR #920, opened 2026-05-27
Add Error handling for data of invalid array shape for add_slice in Async filewriter 

cappel89:
There was a bug in the NIDAQ backend code from the beamline that showed that we should check the dimensions of the data that we receive, and properly lock the error message in case of data coming with an invalid shape. This PR adds this capabilities.


wakonig (approved):
lgtm

PR #929, opened 2026-06-02
fix: add logging option for directory creation in get_full_path

wakonig:
## Description

`get_full_path` creates the directory by default. To be able to migrate to a less-permissive service account, we need to move the file access to the file writer service. For now, we should get a better awareness of methods that actually create them. To this end, I propose to add a log to the directory creation. We can monitor it later with grafana. 
I would expect that they actually don't create the directory as the file writer should create it as soon as the scan is opened but there could be a timing issue. 

## Type of Change

- Add log to the get_full_path 
- Add a `wait_for_directory` helper method

### DIFF
--- a/bec_lib/bec_lib/file_utils.py
+++ b/bec_lib/bec_lib/file_utils.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import time
 import warnings
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


bec_lib/bec_lib/file_utils.py
```diff
@@ -192,10 +199,33 @@ def get_full_path(scan_status_msg: ScanStatusMessage, name: str, create_dir: boo
     # Compile full file path
     full_path = f"{file_base_path}_{name}.{file_extension}"
     if create_dir:
+        if log_if_dir_does_not_exist and not os.path.exists(os.path.dirname(full_path)):
+            logger.warning(f"Directory {os.path.dirname(full_path)} does not exist. Creating it.")
         os.makedirs(os.path.dirname(full_path), exist_ok=True)
     return full_path
 
 
+def wait_for_directory(path: str, timeout: float = 10.0, interval: float = 0.1) -> None:
```
wyzula-jan, line 208: I see usage only in tests, are there any plans to use it somewhere in production?

wyzula-jan (changes requested):
One minor comment, otherwise good.

bec_lib/bec_lib/file_utils.py
wakonig: If we find it necessary, we can use it in plugin repos. I thought it would be a logical companion function to the "removal" of direct directory creation on the device server.

wyzula-jan: ok

wyzula-jan (approved):
🚀🚀🚀

PR #938, opened 2026-06-05
fix: implement request ID handling for scan abortion

wakonig:
## Description

Pending requests were not properly removed from the queue. This PR fixes it by adding support for aborting element through their request ID. While this would normally not happen, or is at least very unlikely to ever be encountered by a user, the scan interlock exposed this faulty behavior: If the scan interlock is active, any request will remain pending. Therefore, if a user submits a new request and notices that the interlock is active and aborts it e.g with ctrl-c, the request would remain in the queue. 

## Type of Change

- Add support for scan modifications with request ID
- Extend ScanModificationMessage to support request ID
- Trigger aborts for pending requests on KeyboardInterrupt

## How to test

- Run unit tests
- Enable the scan interlock and move it to locked state. Submit a umv command through the cli and ctrl-c it. Check bec.queue to ensure that the element is properly removed from the queue.

## Additional Comments

The scan queue widget in BW is still only using the scan id so it will have the same problem. We should fix it asap.

### DIFF
--- a/bec_ipython_client/bec_ipython_client/callbacks/ipython_live_updates.py
+++ b/bec_ipython_client/bec_ipython_client/callbacks/ipython_live_updates.py
@@ -189,7 +189,7 @@ class IPythonLiveUpdates:
                 request_context.queue_status = queue.status
                 self._request_block_id = req_id = self._active_request.metadata.get("RID")
 
-                while queue.status not in ["COMPLETED", "ABORTED", "HALTED"]:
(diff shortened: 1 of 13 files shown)


d-perl (commented):
can we deprecate the fields we don't want to use anymore (https://pydantic.dev/docs/validation/latest/concepts/fields/#deprecated-fields) (and ignore the warnings inside the scan manager with https://docs.python.org/3/library/warnings.html#warnings.catch_warnings until removal ?)

d-perl (approved):
lgtm

PR #943, opened 2026-06-09
fix: initialize ready state in BeamlineStateManager

wakonig:
## Description

* Fix a small bug in the beamline state manager that would leave the manager in a "not ready" state. 
* Add some more logging around the actors.
* Fix startup state handling: The scan server was reporting "ready" before the states were set up.


d-perl (approved):
As you say I think this is a correct change to make either way

PR #944, opened 2026-06-10
test: increase ramp-up time for delay device in test fixture

wakonig:
## Description

The backend logs suggest that everything is working correctly: The actor kicks in, locks the queue and the device recovers, leading to an unlock. The client in the e2e test seems to receive the data too late, i.e. when the device has already recovered. I suggest to increase the rampup time to give it more time. 

While testing, I also found and fixed a bug in the alarm handler.


d-perl (approved):
this does seem to help, the interlock test ran very early on and still passed

PR #949, opened 2026-06-12
fix: migrate extra fields in pydantic models to json_schema_extra

wakonig:
## Description

Pydantic marks additional kwargs in Field as deprecated and recommends to migrate to `json_schema_extra`.

### DIFF
--- a/bec_lib/bec_lib/bl_states.py
+++ b/bec_lib/bec_lib/bl_states.py
@@ -58,7 +58,7 @@ class BeamlineStateConfig(BaseModel):
     state_type: ClassVar[str] = "BeamlineState"
 
     name: str = Field(
-        **ScanArgument(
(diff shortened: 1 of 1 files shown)


d-perl (changes requested):
I don't think this is exactly right, these are standard kwargs for `Field`, not extra kwargs. The type checker just doesn't know that the stuff from `scan_argument.model_dump()` has the right kwarg names. I'm not sure why it's important to use the `ScanArgument` here but if you want to make that work with typing, maybe make a thing like

```python
class ScanArgument(...):

    def to_field(self, args):
        return Field(args, description=self.description, display_name=self.display_name, json_schema_extra={<stuff which isn't in field args>})
```

and then

```python
name: str = ScanArgument(...).to_field(default)
```
?

PR #951, opened 2026-06-12
fix: skip ScanArgs in schema generation

wakonig:
## Description

Skip ScanArgument in schema generation.


d-perl (approved):
nice, good solution

PR #957, opened 2026-06-15
Feature/async signal index

### LINKED ISSUE #755: [FEAT]: AsyncSignal should push shape information on data to additional Endpoint
Async data is currently pushed to a stream where data expires after 10min. So if tools (e.g. plotting) intend to plot data 10minutes after a scan started, they will not receive all data as its already expired in REDIS. 

Therefore, we need a way to know which data point correlates to the async data in a message. We discussed the following:

- Add new endpoint (list)
- Push shape information and stream-id with same expire policy as for the async data.
- In addition, store that information also in the AsyncSignal to allow devices to efficiently wait for data to be received on the backend side.

wakonig:
## Description

This PR introduces async signal indices for keeping track of individual updates of async signal events. This will become useful once we have the data api and try to match async and monitored signals. 

## Related Issues

closes #755

## Type of Change

- new endpoint `device_async_signal_index`
- new message type `DeviceAsyncSignalIndexMessage`
- bec message handler calculates the index and pushes the data to a list in redis

## How to test

- Run unit tests
- Check redis for the new endpoint

### DIFF
--- a/bec_lib/bec_lib/messages.py
+++ b/bec_lib/bec_lib/messages.py
@@ -689,6 +689,36 @@ class DeviceMessage(BECMessage):
         return v
 
 
+class DeviceAsyncSignalIndexMessage(BECMessage):
+    """Message type for sending device async signal index information from the device server
+
+    This message is used to send index information for async signals emitted by the device server. It is published to a
+    separate Redis list and is meant as a companion message to the DeviceMessage that contains the async signal data.
+    The DeviceAsyncSignalIndexMessage contains information about the shapes of the async signals
+    that are being emitted and can be used to keep track of offsets, even after the raw data stream has already been
+    evicted from Redis. The indices are managed on the device server side and are incremented for each new async signal emitted.
+
+    Args:
+        scan_id (str): Unique scan ID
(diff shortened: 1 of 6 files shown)


bec_lib/bec_lib/messages.py
```diff
@@ -611,6 +611,32 @@ def check_metadata(cls, v):
...
+        shapes (dict[str, list[int]]): Dictionary containing the shapes of the async signals being emitted, keyed by signal name
+        indices (dict[str, int]): Dictionary containing the current indices of the async signals being emitted, keyed by signal name.
+                                    The index starts at 0 and is incremented for each new async signal emitted and can be used to
+                                    keep track of offsets in the raw data stream.
+        async_update (DeviceAsyncUpdate): Metadata controlling how data is aggregated into datasets during a scan. See DeviceAsyncUpdate for details.
+    """
+
+    msg_type: ClassVar[str] = "device_async_signal_index"
+
+    scan_id: str
```
cappel89, line 714: add device :str

wakonig: done

cappel89 (approved):

PR #963, opened 2026-06-17
refactor: modularise redis connector

d-perl:
Split RedisConnector into HLI and internal functionality for a cleaner separation between higher order business logic and "backend" managed redis interaction, as preparation for adding a buffer layer to the backend

Needs a small fix in bec widgets, https://github.com/bec-project/bec_widgets/pull/1197

### DIFF
--- a/bec_lib/bec_lib/redis_connector/hli.py
+++ b/bec_lib/bec_lib/redis_connector/hli.py
@@ -1,0 +1,248 @@
+"""
+This module provides a connector to a redis server. It is a wrapper around the
+redis library providing a simple interface to send and receive messages from a
+redis server.
+"""
+
+from __future__ import annotations
+
+import traceback
+
+from redis.client import Redis
+
+from bec_lib.logger import bec_logger
+
+from .buffered_redis_connector import BufferedRedisConnector
+
+logger = bec_logger.logger
+
+
+class RedisConnector:
+    connector_cls = BufferedRedisConnector
+
+    def __init__(
+        self,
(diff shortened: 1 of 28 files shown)


bec_lib/bec_lib/redis_connector/hli.py
```diff
@@ -0,0 +1,435 @@
...
+        """
+        Authenticate to the redis server.
+        Please note that the arguments are keyword-only. This is to avoid confusion as the
+        underlying redis library accepts the password as the first argument.
+
+        Args:
+            username (str, optional): username. Defaults to "default".
+            password (str, optional): password. Defaults to "null".
+        """
+        return self._buffered_connection.authenticate(username=username, password=password)
```
wakonig: I like the refactoring a lot! I'm only wondering if there is a better name for _buffered_connection. In the end, the buffering is likely configurable so the name could be somewhat misleading.

bec_lib/bec_lib/redis_connector/managed_redis_connection.py
```diff
@@ -16,397 +16,47 @@
 import threading
 import time
 import traceback
-import warnings
 from collections.abc import MutableMapping, Sequence
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
-from functools import wraps
```
wakonig, line 23: let's also change the module name from buffered_redis_connector to managed_redis_connection

wakonig (approved):

PR #974, opened 2026-06-23
fix: first axis should be the fast axis

wakonig:
## Description

This PR fixes a bug in particular for the grid scan: The fast axis should be the first one, not the inner one. 

## Type of Change

- Make the first axis always the fast one. This is much easier to communicate and easier to understand when using the GUI. 

## How to test

- Run unit tests
- Plot the positions with BW


bec_server/bec_server/scan_server/scans/position_generators.py
```diff
@@ -327,7 +327,9 @@ def _get_positions_recursively(current_axes):
             positions.extend([[val] + sp for sp in sub_positions])
         return positions
 
-    return np.array(_get_positions_recursively(_axes_arrays))
+    # Build the traversal in reverse axis order so the first user-provided axis is the fast one.
```
cappel89, line 331: This now flis the fast axis to be the first one given to the arguments of the generator right? Let's add this also to the docstring.

cappel89 (approved):

PR #975, opened 2026-06-23
Fix/path optimizer

wakonig:
## Description

This PR fixes the trajectory optimization for the v4 fermat spiral and improved the API. You can now select the column sorting and toggle "snaking". In addition, the default for the column sorting is not ascending as most scans are given with a positive range. This however, breaks a test in cSAXS and needs to be fixed once this PR is merged (https://gitea.psi.ch/bec/csaxs_bec/pulls/237)


cappel89 (approved):
lgtm

PR #991, opened 2026-06-30
Feature/device locks

wakonig:
## Description

This PR adds an ownership model to the device serializer and uses the ownership modes to determine whether a scan should acquire a lock for the duration of the scan. Moreover, only owned devices are used in stateful operations, such as stage, unstage and stop_all_devices. 

These changes are only applied to v4-type scans.

## Definition of Done

### DIFF
--- a/bec_server/bec_server/scan_server/device_lock_registry.py
+++ b/bec_server/bec_server/scan_server/device_lock_registry.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import threading
+import time
 from collections import defaultdict
 from collections.abc import Callable, Iterable
 
@@ -13,6 +14,7 @@ class DeviceLockRegistry:
     """Registry that tracks per-request device locks for the scan server."""
 
     WAIT_INTERVAL_S: float = 0.1
+    WAIT_LOG_INTERVAL_S: float = 5.0
 
     def __init__(self) -> None:
         """Initialize the device lock registry."""
@@ -25,6 +27,7 @@ class DeviceLockRegistry:
         request_id: str,
         devices: Iterable[str],
         interruption_callback: Callable[[], None] | None = None,
+        wait_state_callback: Callable[[list[str]], None] | None = None,
     ) -> list[str]:
         """
         Acquire locks for multiple devices on behalf of a request.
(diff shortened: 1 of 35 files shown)


bec_server/bec_server/scan_server/scans/scan_actions.py
```diff
@@ -202,6 +209,7 @@ def pre_scan(
         if len(device_names) == 0:
             status.set_done()
             return status
+        self.acquire_device_lock(device_names)
```
cappel89, line 236: Shouldn't this be acquired before creating the status

bec_server/bec_server/scan_server/scans/scan_actions.py
```diff
@@ -896,6 +929,47 @@ def rpc_call_no_wait(
         self._send(msg)
         return status
 
+    def acquire_device_lock(
```
cappel89, line 972: Is there a use case for releasing a single device_lock?

bec_server/bec_server/scan_server/device_lock_registry.py
```diff
@@ -0,0 +1,115 @@
...
+                        current_owner,
+                    )
+                    has_logged_wait = True
+
+                self._condition.wait(timeout=self.WAIT_INTERVAL_S)
+
+            if interruption_callback is not None:
+                interruption_callback()
+
+    def release_all(self, request_id: str) -> list[str]:
```
cappel89, line 140: Same comment, any use in releasing a lock for a request_id on a single device, but not all at once? 

PS: If no clear use-case comes up, I would not add it to reduce risks of keeping devices locked by accident.

bec_server/bec_server/scan_server/scans/scan_actions.py
wakonig: I don't think we should really open the door to conditionally acquiring and releasing the locks. This will fire back quite quickly :D

cappel89 (dismissed):
To me this generally looks good. So in principle it feels like we could merge it, but it also feels dangerous without migration to scansv4... On the other hand nobody uses the second queue afaik,..

cappel89 (approved):
PR looks good, but please make an issue that the BW ProgressBar should be tailored to a scan-id. At this point, if you run a scan in the secondary queue, the BW ProgressBar goes nuts between visualizing progress of both scans. Probably it's best to limit the ProgressBar updates to a single queue, or make it configurable through right clicks. Default on primary queue.

PR #994, opened 2026-07-02
feat(scan): v4 scan parity with v3

wakonig:
## Description

Step towards the v4 release. 

## Type of Change

- Fixes a bug in the metadata forwarding of v4-type scans
- Adds e2e parametrization for v3 and v4 scans such that most tests are now executed with both systems
- Adds the missing DeviceRPC scan class and makes it the new default for device rpcs

## How to test

- Run unit tests
- Run rpc calls

## Additional Comments

This could be the last step before the v4 release unless we notice some problems. 

Should be merged after #995


cappel89 (approved):

cappel89 (approved):
lgtm

PR #1007, opened 2026-07-23
feat(device): simplify device info structure by removing redundant keys

wakonig:
## Description

This PR simplifies the device info: https://github.com/bec-project/bec/pull/1007/changes#diff-48a79e17403712240993cc8b6d62287987743c8c3f9c064dbd858f64b491aea4L311

The device name is already part of the device info message so we don't need to add it to the info as well. With the device name gone, there is no need for a nesting. 

## Additional Comments

This should allow us to write a proper device info api more cleanly.

We need to update some tests in BW though.

## Definition of Done

### DIFF
--- a/bec_lib/bec_lib/device.py
+++ b/bec_lib/bec_lib/device.py
@@ -245,7 +245,7 @@ class DeviceBase:
         set_device_config(self, config)
         if info is None:
             info = {}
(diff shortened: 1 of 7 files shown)


cappel89 (approved):
Looks good

PR #1010, opened 2026-07-27
Feature/v4 default scans

wakonig:
## Description

This PR replaces the internal legacy scans with v4 scans and renames them accordingly. The feature parity has already been established with the previous PR (#994). 

## How to test

- Run unit tests
- Run scans, stop them, etc. 

## Potential side effects

We need to adapt the scan control widget in BW to filter the v4 scans to only show relevant ones. Two new scan info flags have been introduced in this PR to accomplish that: "is_internal" and "is_scan". 

## Additional Comments

While it does not remove the ability to run legacy scans, it doesn't expose them anymore. Once the OMNY scans are tested on the new system, we can remove them completely.  

## Definition of Done

### DIFF
--- a/bec_ipython_client/tests/end-2-end/test_scans_e2e.py
+++ b/bec_ipython_client/tests/end-2-end/test_scans_e2e.py
@@ -303,6 +303,7 @@ def test_umv_ctrl_c_stops_motion(bec_ipython_client_fixture: BECIPythonClient):
         dev.samx.velocity.set(original_velocity).wait()
 
 
+@pytest.mark.timeout(100)
 def test_limit_error(bec_ipython_client_fixture):
     bec = bec_ipython_client_fixture
     bec.metadata.update({"unit_test": "test_limit_error"})
--- a/bec_ipython_client/tests/end-2-end/test_scans_e2e.py
+++ b/bec_ipython_client/tests/end-2-end/test_scans_e2e.py
@@ -31,16 +31,13 @@ if TYPE_CHECKING:  # pragma: no cover
 
 
 @pytest.mark.timeout(100)
-@pytest.mark.parametrize("scan_name", ["grid_scan", "_v4_grid_scan"])
-def test_grid_scan(capsys, bec_ipython_client_fixture, scan_name):
+def test_grid_scan(capsys, bec_ipython_client_fixture):
     bec = bec_ipython_client_fixture
     scans = bec.scans
     bec.metadata.update({"unit_test": "test_grid_scan"})
     dev = bec.device_manager.devices
     scans.umv(dev.samx, 0, dev.samy, 0, relative=False)
-    status = getattr(scans, scan_name)(
-        dev.samx, -5, 5, 10, dev.samy, -5, 5, 10, exp_time=0.01, relative=True
-    )
+    status = scans.grid_scan(dev.samx, -5, 5, 10, dev.samy, -5, 5, 10, exp_time=0.01, relative=True)
     assert len(status.scan.live_data) == 100
     assert status.scan.num_points == 100
     captured = capsys.readouterr()
@@ -48,13 +45,12 @@ def test_grid_scan(capsys, bec_ipython_client_fixture, scan_name):
 
 
 @pytest.mark.timeout(100)
-@pytest.mark.parametrize("scan_name", ["fermat_scan", "_v4_fermat_scan"])
-def test_fermat_scan(capsys, bec_ipython_client_fixture, scan_name):
+def test_fermat_scan(capsys, bec_ipython_client_fixture):
     bec = bec_ipython_client_fixture
     scans = bec.scans
     bec.metadata.update({"unit_test": "test_fermat_scan"})
     dev = bec.device_manager.devices
-    status = getattr(scans, scan_name)(
+    status = scans.fermat_scan(
         dev.samx,
         -5,
         5,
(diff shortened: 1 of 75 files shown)


bec_lib/tests/test_devices.py
```diff
@@ -313,12 +313,12 @@ def test_handle_rpc_response_returns_dict(dev: Any):
     assert dev.samx._handle_rpc_response(msg) == {"a": "b"}
 
 
-def test_prepare_rpc_msg_uses_v4_device_rpc_signature(dev: Any):
+def test_prepare_rpc_msg_usesdevice_rpc_signature(dev: Any):
```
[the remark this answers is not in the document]
wyzula-jan: catastrophe

bec_server/bec_server/scan_server/scan_manager.py
```diff
@@ -33,6 +33,15 @@
...
 logger = bec_logger.logger
 
+INTERNAL_SCAN_CLASSES = {
+    "ScanStubs",
+    "ScanComponent",
+    "AsyncFlyScanBase",
+    "SyncFlyScanBase",
+    "ScanBase",
+    "DeviceRpc",
+}
```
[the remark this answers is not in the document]
wakonig: you cannot use it anymore

bec_server/bec_server/scan_server/scan_manager.py
```diff
@@ -165,6 +194,8 @@ def update_available_scans(self, reload: bool = False):
             self.available_scans[scan_cls.scan_name] = {
                 "class": scan_cls.__name__,
                 "base_class": base_cls,
+                "is_scan": scan_cls.is_scan if hasattr(scan_cls, "is_scan") else False,
```
wyzula-jan, line 206: this will be default false for all v3 scans, but even v3 scans are still scans. I think we should flag them as scans but also mark them as internal

bec_ipython_client/tests/end-2-end/test_scans_e2e.py
```diff
@@ -31,30 +31,26 @@
 
 
 @pytest.mark.timeout(100)
-@pytest.mark.parametrize("scan_name", ["grid_scan", "_v4_grid_scan"])
-def test_grid_scan(capsys, bec_ipython_client_fixture, scan_name):
+def test_grid_scan(capsys, bec_ipython_client_fixture):
```
wyzula-jan, line 34: later in this e2e line 806 is DeviceProgressScan tested which is v3 (cannot comment there since no changes were made)

bec_server/bec_server/scan_server/scan_manager.py
```diff
@@ -33,6 +33,15 @@
 
 logger = bec_logger.logger
 
+INTERNAL_SCAN_CLASSES = {
```
wyzula-jan, line 36: why we have flag is_internal when we have a list of internal scans?

wyzula-jan (changes requested):
first round

bec_lib/bec_lib/scans.py
```diff
@@ -133,10 +133,6 @@ def _run(
 
         sys_config = sys_config.model_dump()
         # pylint: disable=protected-access
-        if scans._scan_group:
-            sys_config["queue_group"] = scans._scan_group
-        if scans._scan_def_id:
-            sys_config["scan_def_id"] = scans._scan_def_id
         if scans._dataset_id_on_hold:
             sys_config["dataset_id_on_hold"] = scans._dataset_id_on_hold
```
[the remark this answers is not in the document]
wakonig: yep. on purpose

bec_server/bec_server/scan_server/scan_manager.py
wakonig: fixed

bec_ipython_client/tests/end-2-end/test_scans_e2e.py
wakonig: This is from a plugin repo

bec_server/bec_server/scan_server/scan_manager.py
wyzula-jan, line 206: This way all legacy scans will fallback default to False, since `RequestBase` never defines it. Is it intentional? Even that they are legacy, technically they are scans.

bec_server/bec_server/scan_server/scan_manager.py
```diff
@@ -72,16 +81,35 @@ def get_available_scans() -> list[tuple[str, Type]]:
...
+
+        def _append_new_scan_members(
+            members: list[tuple[str, Type]], candidates: list[tuple[str, Type]]
+        ) -> None:
+            seen_scan_names = {
+                scan_cls.scan_name
+                for _, scan_cls in members
+                if hasattr(scan_cls, "scan_name") and scan_cls.scan_name
+            }
+            for name, scan_cls in candidates:
```
wyzula-jan, line 98: In this decision tree shadowed scans are dropped silently (maybe should be logged). Of course v3 is shadowed by v4 but collisions with for example plugin scans should be IMO logged.

wyzula-jan (changes requested):
minor comments, for logging shadowed scans I have fix which I can push

bec_server/bec_server/scan_server/scan_manager.py
wakonig: but not all of them and we don't have that field. I suggest to only rely on this field for new scans. We are anyway almost done with the migration

wyzula-jan: yeah I think really soon we can drop v3 completely

bec_server/bec_server/scan_server/scan_manager.py
wakonig: I changed the logic now a bit: It logs on duplicated plugins but not on legacy scans.

wyzula-jan (approved):
🚀🚀🚀

PR #1016, opened 2026-07-30
feat: implement client restart functionality and associated messaging

wakonig:
## Description

This PR adds a client restart message and a go lib to emit it. 
When a redeployment is performed, we can use these tools to automatically quite outdated clients. Stale session are quite dangerous and should always be removed. While we do point this out to the users, this is commonly forgotten and it is sometimes difficult to find and remove all stale clients. 

While implementing it, I also refactored the go tools into a more modular lib and added the option to load an acl file from disk. 

## Additional Comments

We will need to implement a similar restart mechanism also for the main app. The companion mode (ipython + companion app) would already shut down properly but the standalone companion app wouldn't.


wyzula-jan (approved):
Lets GO! 🚀

PR #1019, opened 2026-08-06
fix: redis connection init param saving

d-perl:
In the Redis connection initialisation, the connection parameters are backed up in case the authentication fails. This was incompatible with higher versions of redis-py, as more un-deepcopyable parameters were added, similar to the `server` before. This change backs up and restores only the modified parameters, with no need to deepcopy the entire kwargs dict, and thereby adds compatibility for redis-py v8.x


bec_lib/bec_lib/redis_connector/managed_redis_connection.py
```diff
@@ -138,7 +138,14 @@ def authenticate(self, *, username: str = "default", password: str | None = "nul
             self._restart_pubsub()
         except redis.exceptions.RedisError as exc:
             self._redis_conn.connection_pool.reset()
-            self._redis_conn.connection_pool.connection_kwargs.update(old_kwargs)
+            if old_user is _not_set_sentinel:
```
wakonig: Maybe as an alternative: 

```python 
kwargs = self._redis_conn.connection_pool.connection_kwargs

for key, old_value in (("username", old_user), ("password", old_pass)):
    if old_value is _not_set_sentinel:
        kwargs.pop(key, None)
    else:
        kwargs[key] = old_value

```

d-perl: ah yeah, how about like this, no need for the sentinel that way either

wakonig (approved):

PR #1020, opened 2026-08-07
Feat/data api

wyzula-jan:
## Description

This PR introduces the **Data API** (`bec_lib.data_api`) — a unified, client-side data access layer for scan data. It
gives every consumer (plotting widgets, scripts, services) one subscription contract for three kinds of data that
previously required three different code paths: the currently running scan, finished scans from the scan history, and
scan-independent device streams.

### Architecture

**Facade.** `client.data_api.subscribe(sources, scan, callback, ...)` is the single entry point.
`sources` is a list of `(device, entry)` pairs; `scan` selects the scope: `"live"` follows the active scan and every
scan after it, a concrete scan id serves exactly that scan, and `None`
subscribes to scan-less device streams (readbacks, `monitor_1d` monitors, preview signals).

**Immutable columnar updates.** Every delivery is a full-state `SubscriptionUpdate` snapshot, not a delta: consumers
render the newest update and never accumulate data themselves, which makes emission coalescing lossless (at most one
update per `min_emit_interval`, runtime-adjustable).
`update.aligned()` returns equal-length value columns; `update.axis(...)` resolves an x-axis (index / timestamp /
device); `update.reason` distinguishes `"live"`, `"backfill"`, `"rebind"`
and `"history"` deliveries.

**Ordinal alignment.** Points are keyed by an *ordinal* — the scan point id for monitored signals, the per-scan async
message index for async signals, an arrival counter for legacy unindexed devices — instead of arrival order. Late points
fill their hole, repeated points overwrite in place, and one lagging source can never silently shift a curve against
another. Sources are partitioned into *correlation groups* (`scan`, `async:<tag>`, standalone), and only sources within
a group are aligned with each other.

**Plugin routing.** Three data source plugins claim scans by priority: live (10), history (50, file-backed on a worker
thread), device streams (90). A `scan="live"` subscription moves through a defined lifecycle: bind + backfill of
already-recorded points → coalesced live updates →
`"rebind"` on scan change → terminal flush → automatic re-route to the authoritative history snapshot once the file is
published. Subscribing mid-scan therefore never misses data, and the final state is identical to a later history read.

**Bounded resources.** `max_points` caps retention for endless streams; `size_limit_bytes` arms a size gate — the
history plugin estimates a scan's payload from stored metadata *before reading anything*, and an over-limit load is
withheld until `confirm_size()`.

### Additional improvements on top of the core

- **Production hardening**: per-client instance registry with automatic facade re-creation after
  `close()`, unavailable sources retried on the next scan status instead of failing the subscription, async signal-info
  cache invalidation on device config updates.
- **Performance**: incremental sorted columns with cached snapshots (constant-time repeated
  `aligned()` calls); numpy-native bulk ingest for history reads — file columns stay `ndarray`
  end to end, removing multi-second lock holds and O (n) tuple round-trips on multi-million-point scans.
- **Chunked history reads with progress reporting**: large datasets are read in slabs and a
  `progress_callback` receives the load fraction, so GUIs can show a progress bar instead of a frozen window.
- **Deadlock fix**: `subscribe()` vs. the scan-segment dispatcher callback could deadlock on lock order (Data API lock
  vs. scan-manager lock). Entry points now prefetch scan items before taking the API lock; a deterministic regression
  test reproduces the deadlock on the old code.
- **`fix(bec_lib)`**: `ScanHistory` now stores a new entry *before* firing the update event — previously a callback
  querying the history from the event could miss the scan that triggered it (pre-existing issue, surfaced by the history
  re-route).

Documentation (Diátaxis: learn / how-to / reference) is prepared in a bec_docs PR.

## Type of Change

- New feature (`bec_lib.data_api` package)
- Bug fix (`scan_history` update-event ordering)

## How to test

```bash
pytest --random-order bec_lib/tests/test_data_api.py bec_lib/tests/test_data_api_alignment.py \
       bec_lib/tests/test_data_api_history.py bec_lib/tests/test_scan_history.py
```

- End-to-end coverage was extended in `test_scans_lib_e2e.py`.
- Try examples mention in the documentation branch in IPython client
- test with https://github.com/bec-project/bec_widgets/pull/1269, try all plotting widgets there
- The performance of most widgets (especially ones using Curve) is heavily improved

## Potential side effects

- No message formats were changed (`messages.py` untouched) — servers and GUIs remain independently deployable; existing
  consumers of `scan_item` / `scan_history` are unaffected -> for full compatibility of BEC Widgets with DataAPI if we
  decide to deploy BEC first and widgets later this PR has to be merged as well [**add PR**](https://github.com/bec-project/bec_widgets/pull/1273)
- New API surface only; nothing existing was removed. The companion BEC Widgets PR (bec-project/bec_widgets#1269) builds
  on this branch and requires it.

## Additional Comments

The subscription callback may run on background threads (dispatcher / history worker); Qt consumers should go through
the `QtDataSubscription` bridge added in the widgets PR https://github.com/bec-project/bec_widgets/pull/1269.


bec_lib/bec_lib/data_api/data_api.py
```diff
@@ -0,0 +1,654 @@
...
+        # Per-group emission coalescing state.
+        self._emit_state: dict[str, dict] = {}
+
+        self._relays: list[_WeakEventRelay] = []
+        if self._follow:
+            for event, handler in (
+                ("scan_status", "_on_scan_status"),
+                ("scan_history_update", "_on_scan_history_update"),
+            ):
+                relay = _WeakEventRelay(self, handler)
```
wakonig: I think it would be better to remove the weak ref handling from the data api and instead extend the callback handler in bec lib to support weak refs and pass in a louie.safe_ref instance

PR #1024, opened 2026-08-18
fix(rpc): respect read_only during rpc calls

wakonig:
## Description

Fixing a bug that allowed rpc calls to bypass the read-only contract


d-perl (approved):
Thank you!

PR #1026, opened 2026-08-18
feat(async writer): add tracking for written async signals

wakonig:
## Description

Forward the information about the written async signals from the async writer to the nexus declaration. This allows users to customize their file writer depending on the availability of async signals without checking the nexus file.


d-perl (approved):
lgtm :)

PR #1027, opened 2026-08-18
fix(device): handle KeyboardInterrupt gracefully in RPC calls

wakonig:
## Description

Signals do not implement .stop so we should only call it if it exists.


d-perl (approved):
Surprising that this only turned up now!

PR #1029, opened 2026-08-21
fix: improve error handling and logging for Redis stream connections

wakonig:
## Description

Streams were logging at 1 Hz if the connection got lost. While this is unlikely to happen in production, the development machines may not be able to connect to Atlas from the outside of PSI, leading to massive log file. 
Here, I'm changing the logging behavior to log once on connection loss and once when the connection is re-established. 
The pub sub was already only logging once but was lacking the recovery message, which I added now as well.

### DIFF
--- a/bec_lib/bec_lib/redis_connector/managed_redis_connection.py
+++ b/bec_lib/bec_lib/redis_connector/managed_redis_connection.py
@@ -493,12 +493,13 @@ class ManagedRedisConnection:
             else:
                 return self._redis_conn.xread(topics_ids, block=200) or []  # type: ignore strs are fine key and id types
         except redis.exceptions.ConnectionError:
-            logger.error(self.connection_error_str)
+            return None
         except redis.exceptions.NoPermissionError:
             logger.error(f"Permission denied for stream topics: {set(topics_ids.keys())}")
         # pylint: disable=broad-except
         except Exception:
             sys.excepthook(*sys.exc_info())  # type: ignore # inside except
+        return None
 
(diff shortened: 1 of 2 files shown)


wakonig:
## Focused review: approve

Reviewed commit `35c1261c378f7755f10f5b3fe8f8f4761c62f090`, covering all three changed files (+258/−22). No introduced or exposed defects found. My recommendation is to approve this PR.

### Validation

- **108 tests passed** with randomized order (`--random-order --random-order-seed=1029`) across `test_managed_redis_connection.py`, `test_redis_connector_fakeredis.py`, and `test_stream_subs.py`, using the PR source in an isolated worktree.
- **Two real Redis outage/recovery cycles passed** against a temporary Redis server: each stream/pubsub listener logged once per outage and once on recovery, and message delivery resumed after both restarts.
- A partial history-read failure preserved pending subscriptions and cursors until a successful retry.

### Findings

No blocking findings or cleanup/convention findings against this change.

One pre-existing, nonblocking limitation: timeouts and other generic exceptions still invoke `sys.excepthook` on every retry, so repeated tracebacks remain possible. The outage logging claim is verified for `ConnectionError`; this PR does not eliminate traceback repetition for every failure type.

### What was checked and found fine

- Both production callers handle the changed stream-read return contract.
- Failed reads preserve cursors; history migration preserves message order.
- Stream recovery is not announced merely because subscriptions were removed.
- Retry waits release subscription locks and remain interruptible.
- The replacement listener in the updated fakeredis test is owned by the connector and cleaned up by its fixture.

This was a review only; no source changes were made.

